PR #60, opened 2014-12-31
chore(build): Improved performance of watch and build tasks

nchapman:
Watch is now configured to call specific tasks when certain files change. This will be more work to maintain but the performance should be much better as the project grows. The `requirejs` task now has `development` and `production` targets to improve performance when using `grunt serve`.

@6a68 @pdehaan r?

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ To fetch dependencies and get cooking:
 | Name | Description |
 |------|-------------|
 | `autoprefixer` | Adds vendor prefixes to CSS files based on <http://caniuse.com> statistics.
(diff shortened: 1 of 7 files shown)


pdehaan:
Looks solid to me. Does a `grunt requirejs`, `grunt requirejs:development`, and `grunt requirejs:production` successfully.

Invalid targets (such as `grunt requirejs:dev`) blow up, as expected:
```
$ grunt requirejs:dev

Running "requirejs:dev" (requirejs) task
Verifying property requirejs.dev exists in config...ERROR
>> Unable to process task.
Warning: Required config property "requirejs.dev" missing. Use --force to continue.

Aborted due to warnings.
```

PR #63, opened 2015-01-03
fix: Updating the copyright year and adding repo link

### LINKED ISSUE #62: Another year, another copyright update
In [/app/index.html](https://github.com/mozilla/chronicle/blob/master/app/index.html) we're hardcoding "2014".

Not sure if we want to just do a quick PR update every year to bump the year manually, or if we inject the year value via JavaScript.

pdehaan:
Also added a missing `</div>` so that we pass a basic HTML lint.

Fixes #62

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -23,8 +23,9 @@
     </div>
     <div id="bottom-bar">
       <div class="container">
-        <p class="l-text-center">&copy; 2014 &bull; Mozilla</p>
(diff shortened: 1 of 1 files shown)


pdehaan:
r? @nchapman @6a68 

The footer link is kinda "default" looking, but maybe we can beg the talented Mr @johngruen to help us with some link styling in a future PR.

PR #66, opened 2015-01-05
chore(build): Changing the default build directory to be public/ instead of dist/

### LINKED ISSUE #67: Change default config to output to a public/ directory instead of using dist/
(so says @nchapman, so shall it be done)

pdehaan:
Note that I tweaked the _config/local.json.example_ template, so you'll need to manually tweak your _config/local.json_ config files to match (if you want to see the magic files output to "public/" or somewhere custom).

r? @nchapman @6a68 

Fixes #67


grunttasks/autoprefixer.js
```diff
@@ -5,10 +5,14 @@
 module.exports = function (grunt) {
   'use strict';
 
+  var config = require('../server/config');
+
+  var staticPath = config.get('server.staticPath');
```
nchapman: It'd be nice if the config options we cared about were loaded and attached to the `grunt` object so that each task didn't have to worry about loading the config. What do you think?

nchapman:
We should remove `dist` from the main `.gitignore` and add `*` to the new `public/.gitignore`.

pdehaan:
@nchapman Better like this?

nchapman:
@pdehaan looking good! Was the `app/images/loads-logo.png` meant to be in this last commit?

pdehaan:
No, @nchapman, no it was not. Fixing now.

:sigh:

PR #70, opened 2015-01-06
chore(build): Updating a couple dependencies and removing jshint devDep

pdehaan:
Fixed a couple outdated dependencies and removed an odd "jshint" dependency (odd because that should be included via grunt-contrib-jshint anyways).

We should still have about 2 outdated modules (`mozlog` and `mysql`, both with patch level version bumps -- not that we really care right now):

```
$ npm run outdated

> mozilla-chronicle@0.1.0 outdated /Users/pdehaan/dev/github/chronicle
> npm outdated --depth 0

Package  Current  Wanted  Latest  Location
mozlog     1.0.0   1.0.0   1.0.1  mozlog
mysql      2.5.3   2.5.3   2.5.4  mysql
```

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4,42 +4,42 @@
   "dependencies": {
     "bell": {
       "version": "2.0.0",
-      "from": "https://registry.npmjs.org/bell/-/bell-2.0.0.tgz",
+      "from": "bell@2.0.0",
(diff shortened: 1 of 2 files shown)


nchapman:
Any reason not to bump the other two?

nchapman:
Merging this one. Let's do a follow up to update those other two packages. Doesn't seem like anything critical.

PR #71, opened 2015-01-06
chore(build): Embed config values in compiled JavaScript

### LINKED ISSUE #47: Pass server config into front-end
Seems like a great job for a grunt task.

nchapman:
Fixes #47.

This could end up being brittle, but it does the job for now. We can reassess this if our requirements become more complex.

@pdehaan r?

### DIFF
--- a/app/scripts/config.js
+++ b/app/scripts/config.js
@@ -1,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+], function () {
+  'use strict';
+
+  var config = {
+    oauth: {
+      authEndpoint: '{%= config.get("server.oauth.authEndpoint") %}'
+    }
+  };
+
+  return config;
+});
--- a/grunttasks/template.js
+++ b/grunttasks/template.js
@@ -1,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (grunt) {
+  'use strict';
+
+  var config = require('../server/config');
+
+  // Add custom delimiters since our javascript has both mustache and underscore delimiters embedded
+  grunt.template.addDelimiters('js-config-delimiters', '{%', '%}');
+
+  grunt.config('template', {
+    scripts: {
+      options: {
+        data: {
+          config: config
+        },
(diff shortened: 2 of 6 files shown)


app/scripts/config.js
```diff
@@ -0,0 +1,16 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+], function () {
+  'use strict';
+
+  var Config = {
+    oauth: {
+      authEndpoint: '{%= config.get("server.oauth.authEndpoint") %}'
```
pdehaan: Do we need the entire config object here, or do you just want to pass the `authEndpoint`?

grunttasks/template.js
```diff
@@ -0,0 +1,26 @@
...
+  var config = require('../server/config');
+
+  // Add custom delimiters since our javascript has both mustache and underscore delimiters embedded
+  grunt.template.addDelimiters('js-config-delimiters', '{%', '%}');
+
+  grunt.config('template', {
+    scripts: {
+      options: {
+        data: {
+          config: config
```
pdehaan: Do you want `config: config` here, or just `authEndpoint: config.get('server.oauth.authEndpoint')` and only pass the one value to the template?

Or for that matter, we could just pass the config in via the Gruntfile.js in `initConfig` like we did with the `staticPath`.

app/scripts/config.js
```diff
@@ -0,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+], function () {
+  'use strict';
+
+  var Config = {
```
pdehaan: `Config` or `config`?
We aren't using `new` to create a new instance, so not sure on non-constructor naming conventions. But hey... I just work here.

pdehaan:
Apart from my 1.5 stupid comments, lgtm.

app/scripts/config.js
nchapman: Good point. I'll lowercase it.

grunttasks/template.js
nchapman: I think it's more efficient to make the whole config object available so that we don't have to edit multiple files each time we want to import another config option.

pdehaan: in that case, should i be passing my `staticPath` in via `initConfig` or should i try passing the entire `config` object and then using `'<%= config.get('server.staticPath') %>'` in the individual grunt tasks?

or are we cool with the mild inconsistency?

nchapman: We can try to unify them, but really it's two different cases. The config options we care about are different in grunt than they are in the front-end app, so I think it's all right to have them different.

PR #74, opened 2015-01-07
chore(git): exclude npm-shrinkwrap.json from git diffs

6a68:
This change will exclude npm-shrinkwrap.json from the output of `git diff` as well as `git log -p`.

Got the idea from [SO](http://stackoverflow.com/questions/1016798/excluding-files-from-git-diff).

Thoughts?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+npm-shrinkwrap.json -diff


pdehaan:
:+1: From me. Go for it, unless @nchapman objects in the next... 4 minutes.

6a68:
wow, excellent gif.

PR #75, opened 2015-01-07
feat(front-end): Added basic views for actually viewing visits.

nchapman:
Visualized visits are here!

screen shot 2015-01-07 at 3 15 02 pm

You'll need to have run `server/db/create_test_data.js` to see anything at all.

@pdehaan @6a68 r?

### DIFF
--- a/app/scripts/views/visits/index.js
+++ b/app/scripts/views/visits/index.js
@@ -4,12 +4,27 @@
 
 define([
   'views/base',
-  'stache!templates/visits/index'
-], function (BaseView, VisitsIndexTemplate) {
+  'stache!templates/visits/index',
+  'collections/visits',
+  'views/visits/item'
+], function (BaseView, VisitsIndexTemplate, Visits, VisitsItemView) {
   'use strict';
 
   var VisitsIndexView = BaseView.extend({
     template: VisitsIndexTemplate,
+
+    initialize: function () {
+      this.collection = new Visits();
+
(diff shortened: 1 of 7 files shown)


app/scripts/views/visits/index.js
```diff
@@ -4,12 +4,27 @@
...
+], function (BaseView, VisitsIndexTemplate, Visits, VisitsItemView) {
   'use strict';
 
   var VisitsIndexView = BaseView.extend({
     template: VisitsIndexTemplate,
+
+    initialize: function () {
+      this.collection = new Visits();
+
+      this.listenTo(this.collection, 'add destroy reset', this.render);
```
nchapman: Hmm it might be a bad idea to trigger render on reset and add. I'll look into this.

nchapman: Seems to be a-ok; leaving it as is.

app/scripts/views/visits/item.js
```diff
@@ -0,0 +1,32 @@
...
+    },
+
+    initialize: function () {
+      this.listenTo(this.model, 'change', this.render);
+    },
+
+    destroyModel: function (event) {
+      event.preventDefault();
+
+      if (window.confirm('Destroy this visit?')) {
```
pdehaan: A prettier UX, i have not seen!

nchapman: You know it!

6a68:
awww yissssss

PR #80, opened 2015-01-08
chore(build): Adding grunt-rev and grunt-usemin tasks for asset versioning

pdehaan:


grunttasks/build.js
```diff
@@ -5,18 +5,36 @@
...
+        'css',
+        'useminPrepare',
+        'requirejs:' + target,
+        'template',
+        'rev',
+        'usemin'
+      ];
+    }
+
+    grunt.task.run(taskArr);
```
pdehaan: Not sure if there is a cleaner way to do this nonsense above. It feels a bit too... repeated.

grunttasks/serve.js
```diff
@@ -5,9 +5,15 @@
...
-    'hapi',
-    'watch'
-  ]);
+  grunt.registerTask('serve', 'Builds the dist assets, starts the Hapi server, and watches for changes.', function (target) {
+    if (!target) {
+      target = 'development';
+    }
+
+    grunt.task.run([
+      'build:' + target,
```
pdehaan: This just lets us use `npm start` which proxies `grunt serve` or call `grunt serve:production` directly if you want to build production calibre assets (minified and fingerprinted).

grunttasks/serve.js
```diff
@@ -5,9 +5,15 @@
...
 
-  grunt.registerTask('serve', [
-    'build:development',
-    'hapi',
-    'watch'
-  ]);
+  grunt.registerTask('serve', 'Builds the dist assets, starts the Hapi server, and watches for changes.', function (target) {
+    if (!target) {
+      target = 'development';
+    }
```
pdehaan: Still not sure about this `target` stuff. We already have a `NODE_ENV` flag in [server/config.js](https://github.com/mozilla/chronicle/blob/9e7bd617877a9859937772149545e4c61dc6d9dc/server/config.js#L19) which lets us toggle between 'local', 'awsdev', 'prod', 'test'.

Not sure if it is easier to just use `config.get('env')` -- although then we'd have to potentially handle all 4 environments.

server/routes.js
```diff
@@ -112,12 +112,29 @@ module.exports = [{
...
+      path: path.join(__dirname, '..', config.get('server.staticPath'), 'scripts'),
+      listing: config.get('server.staticDirListing')
+    }
+  }
+}, {
+  method: 'GET',
+  path: '/styles/{param*}',
+  handler: {
+    directory: {
+      path: path.join(__dirname, '..', config.get('server.staticPath'), 'styles'),
```
pdehaan: @nchapman @6a68 -- Is this what you kids were thinking about re: routing? I basically created a whitelist of routes for /images/, /scripts/, and /styles/. Or we could try and create a "*" handler which captures all unmatched routes and points them to `config.get('server.staticPath')` and make sure that 404s still route properly and don't get swallowed by some magic route.

6a68: I'm not totally sure what @nchapman had in mind. To me, it seems cleaner to keep images, scripts, and styles underneath `/public`, and also keeps routing simple.

6a68: @pdehaan to address your question directly, though: I think it's better to whitelist just the routes we intend to support, and keep the routes as explicit and simple as possible.

grunttasks/serve.js
6a68: I don't understand where `target` is getting set, but it seems nice to use `config.get('env')` if we can. Rather than handle all four environments differently, could you just, say, do one thing `if (config.get('env') === 'local')`, and do something else for all other cases?

grunttasks/build.js
6a68: I actually like this because it's incredibly simple. Yeah, it could be fancier, but it's immediately clear what will happen in each branch of the conditional. /me shrugs

server/routes.js
nchapman: This represents an opposing view, but I was just thinking we'd change it to this:

`path: '/{param*}'`

I think we should feel comfortable serving anything from public since it's name kind of implies that it is in fact public. If we do want something more specific, it seems like we could construct one fancy route to cover all the cases.

grunttasks/serve.js
pdehaan: The `target` argument gets set when you call `grunt serve:production` (where `serve` is the "task" and `production` is the "target").
So the logic is saying "if there is no `target` specified" (e.g. `grunt serve`) then set a default value of "development", whereas if the `target` _was_ specified (e.g. `grunt serve:development` or `grunt serve:production`) then use the specified target.

Or something like that.

server/routes.js
pdehaan: OK, I'll play around a bit more with this. I think I tried something similar when we were building the loads front end w/ Hapi+Angular and we saw some issues w/ 404 routing and a generic `/{param*}` route. But my memory is a bit hazy on the details &mdash; so I'll try and build a smaller test case and verify.

nchapman: Thanks Peter!

6a68: @nchapman sure, I'm fine with doing a generic wildcard under `/public`, or maybe you meant putting stuff like `/scripts` at the topmost level? Either way, the API is namespaced under `/v1`, so I don't think the catch-all router should be too complex. @pdehaan whatever works is fine; we'll wind up moving all the assets to some other CDN'd domain eventually anyhow :-)

nchapman:
@pdehaan on it!

grunttasks/rev.js
```diff
@@ -0,0 +1,20 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (grunt) {
+  'use strict';
+
+  grunt.config('rev', {
+    app: {
+      files: {
+        src: [
+          '<%= staticPath %>/bower_components/**/*.js',
```
nchapman: You can remove this one. All bower_components are handled by the requirejs task.

grunttasks/usemin.js
```diff
@@ -0,0 +1,32 @@
...
+    }
+  });
+
+  grunt.config('usemin', {
+    options: {
+      assetsDirs: [
+        '<%= staticPath %>'
+      ]
+    },
+    css: [
```
nchapman: I think you can actually drop this css config section. It should work without it.

app/index.html
```diff
@@ -26,6 +28,8 @@
         <p class="l-text-center">&copy; 2015 &bull; <a href="https://github.com/mozilla/chronicle/" target="_blank">Mozilla</a></p>
       </div>
     </div>
-    <script data-main="assets/scripts/main" src="assets/bower_components/requirejs/require.js"></script>
+
+    <script data-main="scripts/main" src="assets/bower_components/requirejs/require.js"></script>
```
nchapman: You can replace this with the following if you run useminprepare after the requirejs task:

```
<!-- build:js scripts/compiled.js -->
<script data-main="scripts/main" src="assets/bower_components/requirejs/require.js"></script>
<!-- endbuild -->
```

grunttasks/build.js
```diff
@@ -5,18 +5,36 @@
...
+        'template'
+      ];
+    } else {
+      taskArr = [
+        'lint',
+        'clean',
+        'copy',
+        'css',
+        'useminPrepare',
+        'requirejs:' + target,
```
nchapman: This order will get js file revving working:

```js
taskArr = [
  'lint',
  'clean',
  'copy',
  'css',
  'requirejs:' + target,
  'template',
  'useminPrepare',
  'rev',
  'usemin'
];
```

grunttasks/serve.js
```diff
@@ -5,9 +5,15 @@
...
+
+    grunt.task.run([
+      'build:' + target,
+      'hapi',
+      'watch'
```
nchapman: `watch` will need a production build target to work correctly here. It currently uses development.

nchapman:
Looking good @pdehaan! Just one more round of changes :wink:.

nchapman:
Woot! :+1:

PR #85, opened 2015-01-08
chore(build): Updating mozlog and mysql dependencies

pdehaan:
Updating the last 2 outdated modules, as per #70 comment.


nchapman:
:+1:

PR #86, opened 2015-01-09
Fixup auth

### LINKED ISSUE #45: figure out session / nonce story for oauth flow
- we're starting with the redirect flow
- we'll probably need to ditch the relier library, because we want the nonce to be signed on the server, hence, opaque to the client

### LINKED ISSUE #76: Get auth sorted out, working with visits api
The current auth code basically duplicates the hapi bell plugin inside the `/auth/complete` route handler. Decide if it's even worth bothering with bell; if it's simpler to just make it a fully custom auth strategy, do that.

6a68:
I tried to be a bit more descriptive in the commit messages, as it was a little late to make this stuff a set of atomic commits, but I still wanted to capture some extra detail. I'll just paste those in as a PR description.

When doing local development with the test user, this patch lets you simply surf to /auth/login and a session cookie will be set, skipping the oauth flow completely. All other funky special casing code is gone. Dang, that's nice. (Normally, hitting /auth/login redirects to another endpoint, which redirects to the fxa oauth flow, which eventually ends with a cookie being set, if the oauth stuff worked out successfully.)

Comments welcome!

-----

 chore(server): fixup auth

Correctly use Bell to handle the OAuth nonce. Correctly set and validate
cookies (both that the fxaId exists, and that any TTL hasn't expired).
Fixes #45.

Replace log.trace with log.verbose everywhere, since we never really
want a trace of the function calls.

Enable cookie-based sessions for the visits API, and modify the cookie
validation function to always allow access if the `testUser` config
option is set. Fixes #76.

Add crude placeholder API docs for auth endpoints.

-----

 chore(api): simplify handling of test user

With this patch, we handle the test user by simply setting a cookie as soon as
the user hits the /auth/login endpoint. This is the only bit of special case
code that we need; the cookie is otherwise normal, is validated normally, and
none of the affected routes need any special case code. w00t!

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -2,6 +2,21 @@
 
 TODO: formally document supported HTTP status codes, format of error and success responses, HTTPS cookie requirements, and which requirements are relaxed when developing locally.
 
+### Auth API
+
+Chronicle uses Firefox Accounts to log users in, but manages its own session duration.
+
+The session cookie is encrypted server-side, so it is not accessible to the client-side code.
+
+The Firefox Accounts OAuth flow is handled by the server; the front-end just needs to redirect logged-out users to the `/auth/login` endpoint.
+
+Note, if the `testUser` config option is set, the cookie will be created directly by visiting `/auth/login`, ignoring the OAuth step.
+
+### Auth API Methods
+
+- Start the login flow: `GET /auth/login`
+- Logout: `GET /auth/logout`
+
 ### Visits API
 
 A `Visit` is a representation of a user's visit to a website at a particular time. Endpoints that return multiple `Visits` return them as simple lists (arrays) of `Visit` objects.
@@ -35,7 +50,7 @@ Note that the `Visit` object returned to a given user never contains that user's
   - This is the format returned by `Date.toJSON()` in modern browsers.
   - See notes section below for a discussion of our choice of timestamp format.
 
-### API Methods
+### Visits API Methods
 
 Note: Partial updates via `PATCH` are not currently supported, but we can add support later to optimize network traffic.
 
--- a/server/bell_oauth_profile.js
+++ b/server/bell_oauth_profile.js
@@ -1,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+var log = require('./logger')('server.bell.profile');
+var db = require('./db/db');
+var config = require('./config');
+
+// this is the custom provider profile function used by Bell to allow us
+// to convert oauth tokens into profile information.
+function profile (credentials, params, get, profileCb) {
+  log.verbose('obtained oauth tokens: ' + JSON.stringify(credentials));
+  var headers = { headers: {'authorization': 'Bearer ' + params.access_token} };
+  get(config.get('server.oauth.profileEndpoint'), headers, function(data) {
+    // Bell returns the parsed data and handles errors internally
+    log.verbose('exchanged tokens for profile data:' + JSON.stringify(data));
+    // TODO use Joi to validate `data` before sending to DB
+    db.getUserById(data.uid, function (err, user) {
+      if (err) {
+        log.warn('getUserById failed: ' + err);
+        return profileCb('getUserById failed: ' + err);
+      }
+      if (user) {
+        log.verbose('user exists! updating oauth token and setting session cookie...');
+        db.updateUser(data.uid, data.email, params.access_token, function(err) {
+          if (err) {
+            log.warn('updateUser failed: ' + err);
+            return profileCb('logging in user failed: ' + err);
+          }
+          // finally, set the cookie and redirect.
+          log.verbose('logged in existing user ' + data.email);
+          credentials.profile = {fxaId: data.uid};
+          return profileCb(null, credentials);
+        });
+      } else {
+        log.verbose('new user! creating record and setting session cookie...');
+        db.createUser(data.uid, data.email, params.access_token, function(err) {
+          if (err) {
+            log.warn('user creation failed: ' + err);
+            return profileCb('creating new user failed: ' + err);
+          }
+          log.verbose('created new user ' + data.email);
+          credentials.profile = {fxaId: data.uid, isNewUser: true};
+          return profileCb(null, credentials);
+        });
+      }
+    });
+  });
+}
+module.exports = profile;
--- a/server/index.js
+++ b/server/index.js
@@ -11,7 +11,6 @@ var config = require('./config');
 var log = require('./logger')('server.index');
 var authProfileCb = require('./bell_oauth_profile');
 var routes = require('./routes');
-var visits = require('./visits');
 var serverConfig = {};
 var db = require('./db/db');
 
@@ -85,6 +84,5 @@ server.register([
 
 // register routes
 server.route(routes);
-server.route(visits);
 
 module.exports = server;
--- a/server/index.js
+++ b/server/index.js
@@ -48,11 +48,6 @@ server.register([
       log.verbose('inside hapi-auth-cookie validateFunc.');
       log.verbose('session is ' + JSON.stringify(session));
 
-      // special case the situation where test user is enabled
-      if (config.get('testUser.enabled')) {
-        return cb(null, true, config.get('testUser.id'));
-      }
-
       var ttl = config.get('server.session.duration');
       if (ttl > 0 && new Date() > new Date(session.expiresAt)) {
         log.verbose('cookie is expired.');
--- a/server/index.js
+++ b/server/index.js
@@ -5,11 +5,15 @@
 'use strict';
 
 var Hapi = require('hapi');
+var HapiAuthCookie = require('hapi-auth-cookie');
+var Bell = require('bell');
 var config = require('./config');
 var log = require('./logger')('server.index');
+var authProfileCb = require('./bell_oauth_profile');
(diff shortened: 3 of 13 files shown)


server/bell_oauth_profile.js
```diff
@@ -0,0 +1,52 @@
...
+
+'use strict';
+
+var log = require('./logger')('server.bell.profile');
+var db = require('./db/db');
+var config = require('./config');
+
+// this is the custom provider profile function used by Bell to allow us
+// to convert oauth tokens into profile information.
+function profile (credentials, params, get, profileCb) {
```
6a68: The general server-side oauth flow is this:
- user surfs to `/auth/login`, which redirects to `/auth/complete` (because Bell expects only one endpoint for the whole oauth dance, I dunno why)
- bell then takes over, generates an oauth nonce, and redirects the user to fxa oauth endpoint
- the user logs in, fxa redirects back to `/auth/complete`
- bell validates the nonce
- bell then calls this `profile` function, which is supposed to take a valid oauth token, and ask some other endpoint for profile information (the outgoing request is made by the `get` function at line 16)
- once `get` returns profile info, we see if the user exists, and either create or update the user db,
- then call profileCb, which passes the fxa ID back to the `/auth/complete` route handler
- the `/auth/complete` handler inserts that ID into a session cookie.

Feel free to ask if part of this seems unclear.

server/index.js
```diff
@@ -31,11 +38,32 @@ server.register([require('hapi-auth-cookie'), require('bell')], function (err) {
...
+
+      var ttl = config.get('server.session.duration');
+      if (ttl > 0 && new Date() > new Date(session.expiresAt)) {
+        log.verbose('cookie is expired.');
+        return cb(null, false);
+      }
+      log.verbose('cookie is not expired.');
+
+      var fxaId = session.fxaId;
+      db.getUserById(fxaId, function(err, result) {
```
6a68: Note, this function validates the cookie by checking that the fxa ID is in the user DB every time a logged-in page is requested. I think this is ok temporarily, and we can address performance bottlenecks as they arise, but I'm open to taking a different approach if this causes great consternation and wringing of hands.

server/visits.js
```diff
@@ -13,19 +13,13 @@ var config = require('./config');
...
 // TODO: add additional metadata fields to the visit datatype
 // TODO: normalize URLs
-// TODO actually create a session for the fake user
 // TODO: visits pagination
 
 var routes = [{
   method: 'GET',
   path: '/v1/visits',
   config: {
-    // TODO allowing unauthenticated access temporarily
```
6a68: all these changes are just removing funky `testUser` session special-case behavior, yay

docs/API.md
```diff
@@ -2,6 +2,21 @@
...
+
+The Firefox Accounts OAuth flow is handled by the server; the front-end just needs to redirect logged-out users to the `/auth/login` endpoint.
+
+Note, if the `testUser` config option is set, the cookie will be created directly by visiting `/auth/login`, ignoring the oauth step.
+
+### Auth API Methods
+
+- Start the login flow: `GET /auth/login`
+- Logout: `GET /auth/logout`
+
```
6a68: I was pretty tired by the time I got around to adding the docs, so they are looking pretty thin. I'm happy to add more docs here, now or later; comments welcome.

docs/API.md
```diff
@@ -2,6 +2,21 @@
...
 
+### Auth API
+
+Chronicle uses Firefox Accounts to log users in, but manages its own session duration.
+
+The session cookie is encrypted server-side, so it is not accessible to the client-side code.
+
+The Firefox Accounts OAuth flow is handled by the server; the front-end just needs to redirect logged-out users to the `/auth/login` endpoint.
+
+Note, if the `testUser` config option is set, the cookie will be created directly by visiting `/auth/login`, ignoring the oauth step.
```
pdehaan: Probably "OAuth" here for consistency.

server/routes.js
```diff
@@ -21,28 +18,69 @@ module.exports = [{
...
       reply.file(path.join(STATIC_PATH, page));
     }
   }
 }, {
   method: 'GET',
-  path: '/auth/logout',
-  handler: function (request, reply) {
-    request.auth.session.clear();
-    return reply.redirect('/');
+  path: '/auth/login',
```
pdehaan: Totally unrelated to everything, but I like to go off topic...

We already separate routes between _server/routes.js_ and _server/visits.js_. Not sure if it makes sense to abstract the /auth/* routes into something like _server/auth.js_ (or maybe move all the route files into a subdir like _server/routes/auth.js_).

server/routes.js
```diff
@@ -50,67 +88,21 @@ module.exports = [{
...
-                request.auth.session.set({fxaId: payload.uid});
-                reply.redirect('/');
-              });
-            }
-          );
-        }
-      );
+      var duration = config.get('server.session.duration');
+      if (duration > 0) {
+        session.expiresAt = new Date(new Date().getTime() + duration);
```
pdehaan: That's a lot of Dates(). Not sure if we can cheat and use `Date.now() + duration`, or `new Date(Date.now() + duration)`.

pdehaan:
/summon resident oauth/hapi expert @seanmonstar.

Seanmonstar, can you take a look at this PR?

server/bell_oauth_profile.js
```diff
@@ -0,0 +1,52 @@
...
+  log.verbose('obtained oauth tokens: ' + JSON.stringify(credentials));
+  var headers = { headers: {'authorization': 'Bearer ' + params.access_token} };
+  get(config.get('server.oauth.profileEndpoint'), headers, function(data) {
+    // Bell returns the parsed data and handles errors internally
+    log.verbose('exchanged tokens for profile data:' + JSON.stringify(data));
+    // TODO use Joi to validate `data` before sending to DB
+    db.getUserById(data.uid, function (err, user) {
+      if (err) {
+        log.warn('getUserById failed: ' + err);
```
seanmonstar: Some `mozlog` nits: 

1. The first argument should be a String with no spaces. It will become part of the `op` that Heka filters by. You probably won't be logging the verbose lines, but this warn (and the others) will trip things up. You can set `debug` to true in mozlog config to add an assertion you're playing by the rules.
2. If you pass an `Error` as an argument, instead of concatenating, you'll get a stack trace. *shrug*

docs/API.md
nchapman: I think later is fine. It's definitely enough to get going and good placeholders to fill in later :).

server/routes.js
6a68: I was thinking the same thing, but never got around to it. I'll make that change.

server/bell_oauth_profile.js
nchapman: > Bell expects only one endpoint for the whole oauth dance

In that case should we just have the endpoint always be `/auth/login` and change it in fxa?

server/routes.js
nchapman: +1

server/routes.js
nchapman: I'm not sure I completely understand what's happening here. Maybe a few comments would help?

server/index.js
nchapman: Is it worth adding an `isExistingUserById()` function (or some such) to `db`?

server/bell_oauth_profile.js
6a68: @seanmonstar Cool! Thanks for the feedback. I haven't had a chance to run this through Heka yet, great to catch these problems before we go to production.

Does the `var log = require('./logger')('server.bell.profile');` on line 7 handle the initial string with no spaces bit, or is that a different part of the logline namespacing?

server/routes.js
6a68: /me shrugs

nchapman: > /me shrugs

Haha. I just came back to it and read it again and it makes sense now. I misread what duration stood for there. Got it now :+1:

server/bell_oauth_profile.js
6a68: We could, although it doesn't matter a lot to me. If you feel strongly about it, let me know, and I'll change it here. Otherwise, we can file an issue and followup later.

nchapman: Yeah filing an issue sounds good.

server/index.js
6a68: Good question. I think this works for now, since the DB is doing basically the same work in either case, but maybe I'll adjust the DB API when we build the `/profile` endpoint. Does that sound good?

docs/API.md
6a68: :ship: :+1:

docs/API.md
6a68: lol, nice catch

server/bell_oauth_profile.js
6a68: filed #87 to follow up on log formatting before we go to prod

6a68:
@dannycoates r?

server/bell_oauth_profile.js
seanmonstar: @6a68 the 2 strings will be concatenated together. So `log.info('checkAuth')` would create the `op` of `chronicle.server.bell.profile.checkAuth`.

server/bell_oauth_profile.js
6a68: Great! Filed #89.

server/bell_oauth_profile.js
6a68: aha! great. added that to the other issue.

6a68:
@dannycoates Actually, this has been open for a while, I'll just land it. But feel free to nitpick future PRs and/or the code currently in the repo :-)

Issue #98, opened 2015-01-10
error creating user table: Error: ER_TOO_LONG_KEY: Specified key was too long; max key length is 767 bytes


pdehaan:
I'm using OSX Yosemite and just did a `$ brew update` and `$ brew install mysql` then `$ mysql.server start`.

Running `$ node server/db/create_db` gives me some odd errors I can't decipher and stalls my Terminal:

```sh
$ node server/db/create_db.js

chronicle.server.db.utils.VERBOSE: pool "connection" event fired, setting timezone
chronicle.server.db.VERBOSE: dropping any existing database
chronicle.server.db.VERBOSE: old database dropped.
chronicle.server.db.VERBOSE: creating chronicle database
chronicle.server.db.VERBOSE: database successfully created
chronicle.server.db.VERBOSE: now using database
chronicle.server.db.VERBOSE: creating user table
chronicle.server.db.WARN: error creating user table: Error: ER_TOO_LONG_KEY: Specified key was too long; max key length is 767 bytes
chronicle.server.db.WARN: something went wrong: Error: ER_TOO_LONG_KEY: Specified key was too long; max key length is 767 bytes
```

Note that the process doesn't end and I have to Ctrl+C to break free.

Not sure why MySQL hates me in my guts, but I think it's somewhere around this line in [/server/db/create_db.js:70](https://github.com/mozilla/chronicle/blob/465f474e9de68c0629ee00681af295667730f711/server/db/create_db.js#L70).
Maybe we move the `pool.end()` call from line 80 into the `createDatabase()` callback function after line 101. If I move that line locally, I still get the errors, but the script ends with this and frees my Terminal up:
```
chronicle.server.db.utils.VERBOSE: received exit signal, closing pool
```

nchapman:
Try brew update and then brew upgrade mysql

pdehaan:
but but but... i _just_ installed mysql...

I'm very suspicious because THIS worked like a champ:
```
'email VARCHAR(128) NOT NULL,' + // was originally VARCHAR(256)
```
I got the idea from [this StackOverflow answer](http://stackoverflow.com/a/20801384) where somebody said that you should switch to 'latin1' instead of 'utf8' since it takes one byte for one character instead of four. Possible BS, but I can probably test the theory by fudging the field length until it passes.
... Well, looks like I can set it to `VARCHAR(255)` and it works, but `VARCHAR(256)` goes :boom:.

```
$ mysql --version
[...]

PR #99, opened 2015-01-10
bug(db): shorten email field by one char to avoid mysql drama

6a68:


6a68:
ehhh, let's just merge and see if it fixes the problem

6a68:
this fixes #98

pdehaan:
FWIW, I think FxA stores the email as 255 chars, per [**mozilla/fxa-auth-db-server**/db/schema/patch-001-002.sql:5-6](https://github.com/mozilla/fxa-auth-db-server/blob/e3bb993730519815e4c13208bca3290bb2552bf7/db/schema/patch-001-002.sql#L5-L6)

> ```sql
normalizedEmail VARCHAR(255) NOT NULL UNIQUE KEY,
email VARCHAR(255) NOT NULL,
```

PR #107, opened 2015-01-13
feat(front-end): Added sign in and sign out links

### LINKED ISSUE #69: Add ability to sign in/up from the front-end

nchapman:
We'll have to figure out the JS story on the home page, but this is the simplest thing that works for now.

@6a68 r?

Fixes #69


6a68:
@nchapman very nice! works for me, merging

PR #109, opened 2015-01-13
fix(server): strip trailing slashes from server routes

### LINKED ISSUE #73: strip out trailing slashes
hapi is fairly stupid about mapping `/v1/visits/` to `/v1/visits`. Build some middleware to handle this before routing.

pdehaan:
Turned out to be easier than thought, as the paths didn't have any existing trailing slashes; so just a quick config change.

FIxes #73

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -11,18 +11,26 @@ var config = require('./config');
 var log = require('./logger')('server.index');
 var authProfileCb = require('./bell_oauth_profile');
 var routes = require('./routes');
-var serverConfig = {};
 var db = require('./db/db');
+
+var serverConfig = {};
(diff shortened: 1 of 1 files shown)


6a68:
@pdehaan Awesome, thanks!

As far as the stylistic refactorings in this patch, I personally would suggest that your time would be better spent quickly writing some poorly-formatted new code that fixes a bug or adds a feature; that said, code cleanup is also a good thing, and it's ultimately your call how you'd like to contribute.

In the future, I do think it would be good to put stylistic refactorings in separate commits from code changes that alter how code should work.

Thanks again! :+1:

PR #114, opened 2015-01-15
bug(server): emit 401 on auth failure, not 302

### LINKED ISSUE #106: API end points should return 401 instead of 302 when not authorized
Right now API endpoints return a `302` and redirect to root when the user isn't authenticated. Instead they should return a `401` and the front-end should handle directing the user to the authentication page.

6a68:
Fixes #106


6a68:
meh, feeling trigger happy, let's DO THIS

PR #120, opened 2015-01-17
refactor(db): creates the server/db/create_db.sh from a template

### LINKED ISSUE #116: clean up create_db.sh
- maybe insert vars via templating engine?
- maybe convert to node file and exec() each of the psql incantations?
- see also conversation in this PR: https://github.com/mozilla/chronicle/pull/113#discussion_r22983032

pdehaan:
Not sure how I feel about it, but it seems to "work for me".

The `grunt template` task gets built from `grunt build` so building the server/db/create_db.sh should be fairly automatic.

Fixes #116


6a68:
@pdehaan you should `git rebase -i HEAD~3` then squash those beasts together

nchapman:
This concerns me some because you'd have to run `grunt build` (or `grunt template`) before you could create the database. That doesn't feel quiet right...

6a68:
if we're staying with a bash script, using bash vars seems simpler. If we're not using bash any more, JS seems like a natural choice, since we go back to having exactly one place with config values (convict)

pdehaan:
:sniff:



---

OK, I found this... [**grunt-db-pg**](https://www.npmjs.com/package/grunt-pg-db). I'll see if I can run through all the hoops of reassigning owners and all that jazz. Plus keeps it in a familiar Grunt/Node stack.
Although it still is a Grunt task, so maybe that doesn't fix @nchapman's concerns. I guess we could always cheat and wire it up to a postinstall script or something for local dev.

Digging deeper into that, it seems to be based on a decent package, [**pg-db**](https://www.npmjs.com/package/pg-db).

> Helper module atop node-postgres that adds transaction management, simpler query API, event hooks, and more.

Features:
- Automatically return connections to the pool - no need to call `done()`
- SQL errors automatically destroy the connection - no need to call `done(err)`
- Named parameters - ... `WHERE foo = :foo` instead of ... `WHERE foo = $1`

The named parameters may be a lot easier than dealing with indexes all over the place like $2, $7, etc, but this may commit us more into the Postgres stack.

```js
'use strict';

var db = require('pg-db')({
  user: 'chronicle',
  password: 'chronicle',
  host: 'localhost',
  database: 'chronicle'
});

db.query('SELECT * FROM visits WHERE id = :id', {
  id: 'f5b30fdc-0762-47b3-a81d-1a459c90af13'
}, function (err, rows) {
  if (err) {
    return console.error(err);
  }
  console.log('Rows: %s', JSON.stringify(rows, null, 2));
});
```

Doesn't seem to support Promises (based on quick experiments), but trivial to add:

```js
var queryp = Promise.denodeify(db.query);

queryp('SELECT * FROM visits WHERE id = :id', {
  id: 'f5b30fdc-0762-47b3-a81d-1a459c90af13'
}).then(function (rows) {
  console.log('Rows: %s', JSON.stringify(rows, null, 2));
}, function (err) {
  console.error(err);
});
```

pdehaan:
OK, this **grunt-pg-db** seems kind of nice, but verbose.

```js
/* This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */

module.exports = function (grunt) {
  'use strict';

  var config = require('../server/config');

  var dbParams = {
    user: config.get('db.postgres.user'),
    password: config.get('db.postgres.password'),
    host: config.get('db.postgres.host'),
    port: config.get('db.postgres.port'),
    database: config.get('db.postgres.database'),
    ssl: config.get('db.postgres.ssl')
  };

  grunt.config('pgdb', {
    createdb: {
      options: {
        connection: {
          host: dbParams.host,
          user: process.env.USER
        },
        sql: [
          'DROP DATABASE IF EXISTS ' + dbParams.database + ';',
          'REASSIGN OWNED BY ' + dbParams.user + ' TO ' + process.env.USER + ';',
          'DROP USER IF EXISTS ' + dbParams.user + ';',
          'CREATE USER ' + dbParams.user + ' WITH PASSWORD \'' + dbParams.password + '\';',
          'CREATE DATABASE ' + dbParams.database + ' ENCODING \'UTF-8\' LC_COLLATE = \'en_US.UTF-8\' LC_CTYPE = \'en_US.UTF-8\';',
          'GRANT ALL PRIVILEGES ON DATABASE ' + dbParams.database + ' to ' + dbParams.user + ';',
          'ALTER SCHEMA public OWNER TO ' + dbParams.user + ';'
        ]
      }
    },
    createtables: {
      options: {
        connection: dbParams,
        sql: [
          'CREATE TABLE IF NOT EXISTS users (' +
            'fxaId CHAR(32) PRIMARY KEY,' +
            'email VARCHAR(255) NOT NULL,' +
            'oauthToken TEXT,' +
            'createdAt TIMESTAMPTZ(3) NOT NULL,' +
            'updatedAt TIMESTAMPTZ(3)' +
          ');',

          'CREATE TABLE IF NOT EXISTS visits (' +
            'id UUID PRIMARY KEY,' +
            'fxaId CHAR(32) REFERENCES users,' +
            'url TEXT NOT NULL,' +
            'rawUrl TEXT NOT NULL,' +
            'urlHash CHAR(40) NOT NULL,' +
            'title TEXT NOT NULL,' +
            'visitedAt TIMESTAMPTZ(3) NOT NULL,' +
            'updatedAt TIMESTAMPTZ(3)' +
          ');',

          'CREATE UNIQUE INDEX CONCURRENTLY fxaId_visitedAt_id' +
            ' ON visits (fxaId, visitedAt, id);'
        ]
      }
    }
  });
};
```

Not a huge fan of all the interpolation in the `createdb` task SQL, but we may be able to find a way to clean that up. I tried the $1 style example from the docs but was getting errors:
```js
{
  sql: 'CREATE DATABASE $1 WITH TEMPLATE prddb OWNER devs;',
  params: ['tempdb']
}
```

I could probably just use Node's `util.format()` to build the strings, which may be marginally nicer:
```js
sql: [
  format('DROP DATABASE IF EXISTS %s;', dbParams.database),
  format('REASSIGN OWNED BY %s TO %s;', dbParams.user, process.env.USER),
  format('DROP USER IF EXISTS %s;', dbParams.user),
  format('CREATE USER %s WITH PASSWORD \'%s\';', dbParams.user,  dbParams.password),
  format('CREATE DATABASE %s ENCODING \'UTF-8\' LC_COLLATE = \'en_US.UTF-8\' LC_CTYPE = \'en_US.UTF-8\';', dbParams.database),
  format('GRANT ALL PRIVILEGES ON DATABASE %s TO %s;', dbParams.database, dbParams.user),
  format('ALTER SCHEMA public OWNER TO %s;', dbParams.user)
]
```

6a68:
I think #145 is the way to go here. Closing.

pdehaan:
Ha!

Well, technically, the #145 wunder-fix only solves the "what do we do for the 'postgres' username on Travis based runs". It doesn't solve all the issues of every database name, username, password being hard coded into the create_db build script.

6a68:
that's true. I bet we can get by for a while in the current state, though :-P

PR #121, opened 2015-01-17
feat(server): add work queues, create visits using the queue

6a68:
for the moment, the only working job is the one that creates visits.

testing the visit creation with curl:

1.  get the cookies and dump them in a file: `curl -c cookies.txt http://localhost:8080/auth/login`
2. POST using the cookie file, eg: `curl -b cookies.txt -H "Content-Type: application/json" -X POST http://localhost:8080/v1/visits -d '{"url":"bar.com","title":"bar","visitedAt":"2015-01-16T23:41:21.354Z"}'`


server/work-queue/queue.js
```diff
@@ -0,0 +1,106 @@
...
+    log.verbose('*** checked for worker status: ' + verb +
+      ' (event loop delay: ' + delay + 'ms)');
+  });
+  multiWorker.start();
+  log.verbose('queue workers are ready');
+  workersReady.resolve();
+});
+
+// TODO add reject handler when we listen for queue/worker startup failures
+//Q.all([queueReady.promise, workersReady.promise]).then(function() {
```
6a68: http://media.giphy.com/media/3Mm6HZud4m2oU/giphy.gif

config/local.json.example
```diff
@@ -26,13 +26,16 @@
       "app": "chronicle",
       "debug": false,
       "fmt": "pretty",
-      "level": "verbose"
+      "level": "info"
```
6a68: the server is ridiculously chatty on 'verbose', so it probably should not be the default level.

npm-shrinkwrap.json
```diff
@@ -128,6 +128,71 @@
         }
       }
     },
+    "embedly": {
```
6a68: whoops, better remove that...

server/index.js
```diff
@@ -35,6 +37,9 @@ server.connection({
   }
 });
 
+// TODO clean this up, but at least the method is exposed
+server.method('queue.createVisit', createVisit);
```
6a68: I didn't want to `require` the queue more than once, seemed like foul things might occur. Not sure if this is the right way to instantiate it once, though - a plugin seems better. @dannycoates thoughts on the best way to attach a singleton thing (like a work queue) to a hapijs server?

server/work-queue/queue.js
```diff
@@ -0,0 +1,108 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+var nr = require('node-resque');
```
6a68: for right now, most of the contents here are pasted straight out of the node-resque readme. "works on my machine" ;-P

server/work-queue/queue.js
```diff
@@ -0,0 +1,108 @@
...
+
+var connectionDetails = {
+  host: config.get('db.redis.host'),
+  password: config.get('db.redis.password'),
+  port: config.get('db.redis.port'),
+  database: config.get('db.redis.database')
+};
+
+var queueOpts = { connection: connectionDetails };
+var queue = new nr.queue(queueOpts, jobs, function onQueueReady() {
```
pdehaan: ESLint says: "A constructor name should not start with a lowercase letter".
Probably not worth caring about since it seems to object to the "queue" part of it, so we'd have to refactor to something absurd like:
```js
var Queue = nr.queue;
var MultiWorker = nr.multiworker;

var queue = new Queue(queueOpts, jobs, function {...});
```

server/work-queue/queue.js
```diff
@@ -0,0 +1,108 @@
...
+  log.info('queue started');
+  // TODO listen for queue events?
+  queueReady.resolve();
+});
+
+var multiWorkerOpts = {
+  connection: connectionDetails,
+  queues: ['chronicle'],
+  minTaskProcessors: 1,
+  maxTaskProcessors: 10,
```
pdehaan: Trailing comma here.
Plus, not sure if `10` counts as a magic number and should maybe be moved into a Convict config for easy, environment-based tweakage.

nchapman: Yeah convict config sounds good for these.

server/work-queue/queue.js
6a68: ESLint...that's not a linter we currently use, right?

server/work-queue/queue.js
6a68: :+1:

nchapman:
Works for me too! Nicely done! :+1:

PR #122, opened 2015-01-20
refactor(db): switching from  to  module

pdehaan:
DO NOT MERGE

Proof of concept to show switching from [**pg**](https://www.npmjs.com/package/pg) module to [**pg-db**](https://www.npmjs.com/package/pg-db) module.

This lets us remove a bunch of the `pg.connect(dbParams, function(err, client, done) {...});` blocks and take advantage of **pg-db**'s magical abilities to:
> - Automatically return connections to the pool - no need to call `done()`
> - SQL errors automatically destroy the connection - no need to call `done(err)`

This will also make it easier to make a switch from positional arguments (`$1`) to named arguments (`:fxaId`), or if we want to create a promise-based wrapper for queries instead of using callbacks w/ `cb(err, data)`.


server/db/db.js
```diff
@@ -46,50 +42,44 @@ var user = {
...
-      client.query(query, [fxaId, email, oauthToken, new Date().toJSON()], function(err) {
-        if (err) {
-          log.warn('error saving user: ' + err);
-        } else {
-          _verbose('db.user.create succeeded');
-        }
-        done();
-        cb(err);
-      });
+    var args = [fxaId, email, oauthToken, new Date().toJSON()];
```
pdehaan: I left this as `$1`, `$2`...

But we could easily refactor this into something like this:

```js
var query = 'INSERT INTO users (fxaId, email, oauthToken, createdAt) ' +
            'VALUES (:fxaId, :email, :oauthToken, :createdAt)';
var args = {
  fxaId: fxaId,
  email: email,
  oauthToken: oauthToken,
  createdAt: new Date().toJSON()
};
```

server/db/db.js
```diff
@@ -146,34 +132,29 @@ var visit = {
...
-      });
+    var args = [visitedAt, new Date().toJSON(), url, urlHash, url, title, fxaId, visitId];
+
+    db.query(query, args, function(err, r) {
+      if (err) {
+        log.warn('error updating visit: ' + err);
+      } else {
+        _verbose('db.visit.update succeeded');
+      }
+      cb(err, r && visit._normalize(r[0]));
```
pdehaan: So, we may be able to easily switch this to using Promises using something like this:

```js
var queryp = Promise.denodeify(db.query);
// ...
return queryp(query, args).then(function (rows) {
  return visit._normalize(rows[0]);
}, function (err) {
  log.warn('error updating visit: ' + err);
  return err;
});
```

Then a bit of server route tweaking to convert from callbacks to promises...
```js
var onError(err) {
  log.warn(err);
  reply(Boom.create(500));
};
visits.get(fxaId, request.query.count).then(reply, onError);
```

6a68:
I'm not comfortable switching to `pg-db` because it just looks too new / untested: only a couple months old, zero open issues in Github. Sorry, man.

pdehaan:
Current status:

PR #145, opened 2015-01-20
fix(build): Added optional user param to create_db.sh

nchapman:
So here's what I was thinking as a quick fix for specifying the user in the `create_db.sh` script. Good enough for now?

@6a68 @pdehaan r?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ before_script:
   # Copy configuration
   - cp config/local.json.travis config/local.json
   # Setup database
-  - server/db/create_db.sh
+  - server/db/create_db.sh postgres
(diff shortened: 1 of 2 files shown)


6a68:
looks good. how about we add some kind of help text if no args are passed, something like 

```bash
if [ -z "$1" ]
  then
    echo "Please supply a postgres superuser name to run this script, for example, 'create_db someuser'."
    exit 1
fi
```

pdehaan:
If no args are passed, I _believe_ it is just defaulting to `$USER` env var.
Of course, Lorde help you if you're using Windows.

6a68:
doh! you're right. let's merge this puppy

Issue #91, opened 2015-01-09
Get search up and running


6a68:
Yay, elasticsearch! Let's get it running locally and figure out how we'll run it in production. This could just include indexing the title and url.

- [x] Provision elastic search
- [x] Build indexer (worker)
- [x] Add search API
- [x] Fancier second-pass worker / see TODOs in bug comments
- [x] Search results in front-end

6a68:
Just submitted #209 to cover the initial tiny version of search.

I wound up deferring a lot of important stuff until slightly later - another PR, but in this same sprint. Those remaining tasks are:

TODOs:
- connect over ssl? with password?
- migrations / reindexing
  - would be _so nice_ to find an ORM that works with both...I doubt it exists
- pooling ES conns / closing conns on shutdown
  - sadly the ofishul ES client has a big fat TODO in its docs for pooling; maybe use one of those generic node pooling libraries?
- (maybe) split models + db objects, would make the models 95% shorter / easier to parse
- use the `updatedAt` postgres timestamp for ES versioning (optimistic concurrency le sigh)?
  - if so, make it a required field in postgres as well
- figure out how to pass a logger into insane ES client
  - note that ES client has 5 log levels, mozlog 7, and 'warn' != 'warning', fml
- elastic.js for query building? recommended by ES docs, yet looks like abandonware :-\
- add pagination, multi-word query support to ES
  - this can wait until embedly worker lands and we actually have rich data to play with

6a68:
This is definitely fixed, between #209 and #224. w00t!

PR #209, opened 2015-01-22
Issue 91 elasticsearch init

6a68:
This isn't yet integrated into the create_test_data script. However, that test data is not going to search well at all in its current form, since every title is the same.

The best way to play around with this PR is to install the addon, surf around a bit, say, to some mozilla sites, then manually hit the endpoint, something like  `localhost:8080/v1/search?q=mozilla`, and see what you can make out in the returned JSON.

I have a ton of TODOs which I've added to #91, to be handled later in the current sprint. One step at a time :-)


README.md
```diff
@@ -25,9 +25,10 @@ To fetch dependencies and get cooking:
 1. `npm install`
 2. As part of the npm install process, the `postinstall` script will install the Bower dependencies for you.
 3. Copy `config/local.json.example` to `config/local.json`, and put your local info in there.
-4. Run `./server/db/create_db.sh` to create the database, tables, and indexes
+4. Run `./bin/create_db.sh` to create the database, tables, and indexes
   - this script currently hard-codes the db user, password, and dbname to 'chronicle' (issue #112)
-5. Run `node /server/db/create_test_data.js` to create test data (pass that script `--help` for options)
+5. Run `./bin/create_test_data.js` to create test data (pass that script `--help` for options)
```
6a68: doh

6a68:
@nchapman @pdehaan r?

server/models/models.js
```diff
@@ -215,12 +286,56 @@ var visits = {
...
+          // except for downcasing. what's the cleanest way to formalize the
+          // contract between API layers and these 2 databases? it should lead
+          // naturally to API docs, I'd think.
+          // TODO we might also want to return relevance scores or other special
+          // elasticsearch bits as part of this API
+          var s = item._source;
+          return {
+            id: s.id,
+            fxaId: s.fxaId,
+            url: s.url,
```
nchapman: Looks like `title` is missing here.

nchapman:
@6a68 this is looking great. I think I noted it in the correct place above, but the search endpoint isn't returning `title`:

```json
{
  "resultCount": 1,
  "results": [
    {
      "id": "88cfdb5e-c18a-4e58-ba96-7d922c952af4",
      "fxaId": "a6c9411418f94710a712ab97e7a3541c",
      "url": "bar.com",
      "urlHash": "91b6494ac79837919d80c01c89c2e70b4a5479a2",
      "visitedAt": "2015-01-16T23:41:21.354Z"
    }
  ]
}
```

server/models/models.js
```diff
@@ -5,9 +5,22 @@
...
+var log = require('../logger')('server.models');
+
+// >:-(
+// https://github.com/elasticsearch/elasticsearch-js/issues/33
+var esParamsFactory = function() {
+  return {
+    host: {
+      host: config.get('db.elasticsearch.host'),
+      post: config.get('db.elasticsearch.port'),
+      log: 'verbose'
```
pdehaan: Should 'verbose' be hardcoded, or pull it from a convict config?

server/models/models.js
```diff
@@ -133,46 +145,105 @@ var visit = {
...
+          index: 'chronicle',
+          type: 'visits',
+          id: visitId
+        }).then(function() {
+          _verbose('models.visit.delete succeeded');
+          done();
+          cb();
+        }, function(err) {
+          log.warn('error deleting visit in elasticsearch: ' + err);
+          cb(err);
```
pdehaan: Need to call `done()` here?

server/models/models.js
6a68: short answer: this is one of the todos that I added to #91.

longer answer: the elasticsearch people implemented their own logger, which is Bunyan compatible, but which uses 5 loglevels instead of mozlog's 7. They also use "warning" instead of "warn". So, we'll actually need some friggin mapper between the two loggers, or we'll get rid of mozlog. TBD.

server/models/models.js
6a68: yes! great catch :-P

server/models/models.js
6a68: great catch, thank you

6a68:
thanks for the comments, gents

PR #211, opened 2015-01-23
feat(heartbeat): Adding a __heartbeat__ route

### LINKED ISSUE #204: add __heartbeat__ health check endpoint
We really need this before we ask for ops help going to prod.

For right now, let's just detect massive failure:

Cluster is ok if
- webhead's local redis instance is responsive
- postgres and elasticsearch are reachable and responsive

pdehaan:
Fixes #204


pdehaan:
r? @6a68 @nchapman /cc @mostlygeek 

Currently only checks Postgres and ElasticSearch since I didn't see Redis in package.json yet, but looking deeper into the queue, it may be handled in a worker and I may need to dig more.

server/routes/ops.js
```diff
@@ -0,0 +1,75 @@
...
+// https://github.com/elasticsearch/elasticsearch-js/issues/33
+var esParamsFactory = function() {
+  return {
+    host: {
+      host: config.get('db.elasticsearch.host'),
+      post: config.get('db.elasticsearch.port'),
+      log: 'verbose'
+    }
+  };
+};
```
pdehaan: Seems a bit weird that we'll need to use this anti-pattern in multiple files. Not sure if it's worth having some wrapper code that just creates a new `module.exports = new es.Client(esParamsFactory())` for us so all the connection details are in a single place.

server/routes/ops.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+    deferred.resolve();
+  });
+  return deferred.promise;
+};
+
+var esCheck = function () {
+  var deferred = Q.defer();
+  var esClient = new es.Client(esParamsFactory());
+  esClient.search({
```
pdehaan: Not sure if there is a better API to check than `search()`, or if we stick with that since that's a decent test of if something is truly working.

server/routes/ops.js
```diff
@@ -0,0 +1,75 @@
...
+    index: 'chronicle',
```
pdehaan: @6a68 Should these be hardcoded as 'chronicle' or should we have a convict config for this? (I think I just stole the connection code from a models.js file).

server/routes/ops.js
```diff
@@ -0,0 +1,75 @@
...
+    handler: function (request, reply) {
+      Q.all([
+        pgCheck(),
+        esCheck()
+      ]).then(function (results) {
+        reply('ok').code(200);
+      }, function (err) {
+        reply(err.message).code(503);
+      }).catch(function (err) {
+        reply(err.message).code(503);
```
pdehaan: If a promise was `reject()`-ed or error thrown, return an HTTP status code of 503 (server error) instead of vanilla 200.

@mostlygeek Does that seem correct, or do you want some other status code for failures?

server/routes/ops.js
6a68: yeah, I'm abstracting the elasticsearch conncetions into a DBO sorta thing, copypasta is fine for now.

server/routes/ops.js
6a68: for right now, this is fine. once I have a reasonable idea around what the indexes should be, it'll get abstracted out into configs.

server/routes/ops.js
6a68: I was thinking of using `ping`, as in the [example](https://github.com/elasticsearch/elasticsearch-js#examples) in the node elasticsearch library, but `search` is fine for now.

server/routes/ops.js
```diff
@@ -0,0 +1,75 @@
...
+    if (err) {
+      return deferred.reject(err);
+    }
+    deferred.resolve();
+  });
+  return deferred.promise;
+};
+
+var esCheck = function () {
+  var deferred = Q.defer();
```
6a68: the elasticsearch library returns promises, so you don't need all this

server/routes/ops.js
6a68: I'm not sure I understand what the trailing `catch` is doing on line 70. It looks like the idea is to catch a possible error thrown by the `onFulfilled` handler on line 67. But the catch handler is calling the same code as that function, so if it failed on line 67, it'll fail again on line 71, so we don't need it (I think).

It might be better to terminate the chain using `done` in place of `then` on line 66, and skip the trailing `catch`.

server/routes/ops.js
```diff
@@ -0,0 +1,75 @@
...
+
+module.exports = [{
+  method: 'GET',
+  path: '/__heartbeat__',
+  config: {
+    handler: function (request, reply) {
+      Q.all([
+        pgCheck(),
+        esCheck()
+      ]).then(function (results) {
```
6a68: What are the `results`?

server/routes/ops.js
```diff
@@ -0,0 +1,75 @@
...
+        reply('ok').code(200);
```
6a68: I'm not sure what the ideal JSON schema is for the response, but we can tweak it later if we need to.

server/routes/ops.js
6a68: aaaactually, I think I'm looking too hard at this code. It works, it's fine.

server/routes/ops.js
6a68: ...but it works, so it's fine, feel free to leave it

6a68:
@pdehaan This is good to land, except we do need to ping the redis instance that sits underneath `node-resque`. You can use the `redis` library or just curl its endpoint and see if it responds, whatever's easiest.

Feel free to ignore my nitpicking comments...I sometimes get into "this code will live forever / must be perfect" reviewer mode, but really, this is fantastic and plenty readable. Great work! Feel free to merge after adding the redis check.

server/routes/ops.js
6a68: ehh...it's fine, maybe we'll parse `results` at some later point

pdehaan:
Added 'redis' as a dependency in package.json, not sure why Travis is barfing.

package.json
```diff
@@ -19,6 +19,7 @@
     "node-resque": "0.11.5",
     "pg": "4.1.1",
     "q": "1.1.2",
+    "redis": "0.12.1",
```
pdehaan: Had to add this as a dependency for the __heartbeat__ endpoint health check. No big deal since node-resque was including it anyways.

pdehaan:
Alright, alright, alright... Travis now fails, but it looks like it's coming from the `npm test` task and not due to my substandard coding skills.

re-r? @6a68 @nchapman

nchapman:
Well done @pdehaan! :+1:

6a68:
:beers:

PR #215, opened 2015-01-24
fix(db): closing some db pool connections

### LINKED ISSUE #213: A couple of missing done() calls in server model routes
It looks like a few of the routes in /server/models/models.js aren't calling `done()` which probably means something like the database connection into the pool.

1. [/server/models/models.js:171](https://github.com/mozilla/chronicle/blob/a6ee47e6d39d07b67b1b015908c9419afa73127c/server/models/models.js#L171)
2. [/server/models/models.js:215](https://github.com/mozilla/chronicle/blob/a6ee47e6d39d07b67b1b015908c9419afa73127c/server/models/models.js#L215)

pdehaan:
Closes some unclosed Postgres pool connections. A quick search shows 9 instances of `done)` arguments and an equal number of `done()` callback invocations.

> **note:** if you do not call `done()` the client will never be returned to the pool and you will leak clients. This is mega-bad so always call `done()`
> &mdash; via https://github.com/brianc/node-postgres/wiki/pg#connectfunction-callback

Fixes #213

### DIFF
--- a/server/models/models.js
+++ b/server/models/models.js
@@ -84,7 +84,6 @@ var user = {
           _verbose('models.user.get succeeded');
         }
         done();
-
         cb(err, r && user._normalize(r.rows[0]));
       });
     });
@@ -143,9 +142,9 @@ var visit = {
     pg.connect(dbParams, function(err, client, done) {
       if (err) { return onConnectionError(err, cb); }
       client.query(query, [visitId, fxaId, url, url, urlHash, title, visitedAt], function(err, r) {
+        done();
         if (err) {
           log.warn('error creating visit: ' + err);
-          done();
           return cb(err);
         }
         // postgres succeeded, now insert into elasticsearch
@@ -164,7 +163,6 @@ var visit = {
           }
         }).then(function() {
           _verbose('models.visit.create succeeded');
-          done();
           cb();
         }, function (err) {
           log.warn('error adding visit to elasticsearch: ' + err);
@@ -183,9 +181,9 @@ var visit = {
       if (err) { return onConnectionError(err, cb); }
       client.query(query, [visitedAt, updatedAt, url, urlHash, url, title, fxaId, visitId],
         function(err, r) {
+        done();
         if (err) {
           log.warn('error updating visit: ' + err);
-          done();
           return cb(err);
         }
         // postgres succeeded, now update in elasticsearch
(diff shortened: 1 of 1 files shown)


pdehaan:
A counter solution:

Write a `queryp()` function which creates the `pg.connect()` connection, executes the query with specified parameters, auto-closes the pool connection, and returns the query result as a promise.

So something like this:

```js
var sql = 'SELECT id, url, urlHash, title, visitedAt ' +
          'FROM visits WHERE id = $1 AND fxaId = $2';
var args = [visitId, fxaId];
queryp(sql, args).then(function (results) {
  return visit._normalize(results.rows[0]);
}, function (err) {
  log.warn('error getting visit: ' + err);
});
```

Instead of this:
```js
    pg.connect(dbParams, function(err, client, done) {
      if (err) { return onConnectionError(err, cb); }
      client.query(query, [visitId, fxaId], function(err, r) {
        if (err) {
          log.warn('error getting visit: ' + err);
        } else {
          _verbose('models.visit.get succeeded');
        }
        done();
        cb(err, r && visit._normalize(r.rows[0]));
      });
    });
```

pdehaan:
Errr, plus I guess this:

```js
var pg = require('pg');
var config = require('../config');

var dbParams = {
  user: config.get('db.postgres.user'),
  password: config.get('db.postgres.password'),
  host: config.get('db.postgres.host'),
  port: config.get('db.postgres.port'),
  database: config.get('db.postgres.database'),
  ssl: config.get('db.postgres.ssl')
};

module.exports = {
  queryp: function (sql, args) {
    var deferred = Q.defer();
    args = args || [];

    pg.connect(dbParams, function(err, client, done) {
      if (err) {
        return deferred.reject(err);
      }
      client.query(sql, args, function(err, results) {
        done();
        if (err) {
          return deferred.reject(err);
        }
        deferred.resolve(results);
      });
    });
    return deferred.promise;
  }
};
```

**Pros:** Postgres connection config and logic in one place and easily shareable. Don't need to worry about creating or closing a connection when we want to execute a query.
**Cons:** If we have complex routes that require multiple Postgres queries per route, we'd be opening one connection per query instead of potentially reusing an existing client for all queries.

6a68:
@pdehaan Thanks for the patch! This code's all getting rewritten right now, but I don't see any reason we can't land this anyway. Also, I wound up factoring out the postgres connection stuff into a separate class in a branch I've got going; it looks almost exactly like the promise-returning example in your last comment.

PR #224, opened 2015-01-29
Scraper db all the things etc

6a68:
@nchapman @pdehaan ok.....here we go, readyish for review (I need to do a final code cleanup pass, but you guys can help with that :-).

This is a huge diff. I'll make it a goal to do a better job of splitting out subtasks next sprint.

I'm not sure the best way to ask for a review of this code. On one hand, leaving the 30-something individual commits makes it possible to understand wtf is happening in a local sense. On the other hand, some commits undo others, and at the end of the day, the whole massive diff has to make sense together.

The good news is, this is really a lot of fun to play around with, especially if you use the addon-equipped browser to insert some new visits. Suddenly it almost feels like a real product :rainbow:

#### prepare for log spewin' volcano action
You can set the log level to `verbose` to see all the stuff that flows in and out of the work queue, plus a lot of args to the models. This is fun for convincing yourself individual pieces work, but it can also get annoying. Every error is logged at the `warn` level, so setting log config back to `info` will give you errors and the occasional informative logline, and a lot less noise.

#### about the test data and test user
Run create_db and create_test_data to set up the database. create_db now wipes and recreates the chronicle index in elasticsearch; let me know if this fails for either of you. create_test_data now just shoves a list of URLs into the production code path. This means that embedly will scrape those pages in real time; be ready for a LOT of log output if you've got it set to verbose. The create_test_data script bails after 60 seconds, rather than try to figure out some way to quit when the queue is empty (I ran out of steam there).

#### about all the refactoring...
A lot of pieces got split apart, as I've been doing a ton of repl testing to convince myself things work. The route handlers have been moved under /controllers. The db code has been split out from the models. The models have been split into pieces. It's a lot more pieces, but there is some kind of form emerging.

#### adding visits
You should be able to fire up the mozilla-chronicle-addon code, surf to localhost:8080/auth/login, and then surf around and see those visits asynchronously added, then scraped. These new visits should show up in the main page, as well as in the search results.

##### dude, I can't find anything
I know. The current search implementation is horrible. It's only searching the content of the page, not the title or url or anything else that's useful. This makes it challenging to convince yourself a new page has been indexed or has been deleted. We'll fix this soon, maybe before I land this patch.

#### deleting visits
You can still delete visits from the homepage. Again, you should see them get deleted from elasticsearch, and (if you can find the right search term from the article body), you should see them vanish from the search results pretty rapidly.

#### updating visits
I actually haven't verified that this works. Need to dig out my notes on cURLing with cookies. It's not really in the product yet, so seems low priority to me.

#### API docs: what me worry?
I'm returning a crazy mountain of data to the front-end right now. I'm really happy to just see that data, but maybe I should return a saner subset before landing this (thoughts?).

#### not yet handling bulk inserts
Next sprint, I guess.

#### underscores and camelCase
I decided to make life easier by inventing a little convention: all database things are underscored; all JS things are camelCased. The pg layer handles the underscore to camel conversion; all of the queries already refer to underscored versions. Oh, that reminds me:

#### there is yet another doc file
I created server/README halfway through the refactoring, so it's already out of date! *le sigh* I think it's probably best to nuke it, for now, and reinstate it soonish with actually accurate information. I wound up with the models living in a funky limbo, halfway between "model per table" and "model as business logic noun which abstracts multiple tables". I hope to fix that soon - maybe when I take a day to evaluate ORM options. Feedback welcome...I'm happy to see things getting saner, but definitely not yet proud of the code :-P

There are probably a lot of details I'm forgetting. I'll go through the code tomorrow morning and annotate all the failsauce, as well as the winsauce and break-evensauce.


bin/create_db.sh
```diff
@@ -19,22 +19,66 @@ psql -c "CREATE DATABASE chronicle ENCODING 'UTF-8' LC_COLLATE = 'en_US.UTF-8' L
...
+  extracted_provider_display TEXT,
+  extracted_provider_name TEXT,
+  extracted_provider_url VARCHAR(2048),
+  extracted_published TIMESTAMPTZ(3),
+  extracted_safe BOOLEAN,
+  extracted_title TEXT,
+  extracted_type TEXT,
+  extracted_url VARCHAR(2048),
+  created_at TIMESTAMPTZ(3) NOT NULL,
+  updated_at TIMESTAMPTZ(3) NOT NULL
```
pdehaan: This `updated_at` has `NOT NULL` but the other 2 (`users` table and `visits` table) don't.

grunttasks/build.js
```diff
@@ -14,7 +14,6 @@ module.exports = function (grunt) {
 
     if (target === 'development') {
       taskArr = [
-        'lint',
```
pdehaan: :crying_cat_face:

server/config.js
```diff
@@ -81,6 +87,12 @@ var conf = convict({
         doc: 'Require SSL for postgres connections.',
         default: false,
         format: Boolean
+      },
+      queryTimeout: {
+        doc: 'Milliseconds before app-level queries will bail',
+        default: 15000,
+        env: 'PGTIMEOUT',
```
pdehaan: `ES_TIMEOUT` vs `PGTIMEOUT`. Not sure if we should add an underscore or PG or remove it on ES.

6a68: yeah, the PG* env vars are the ones natively supported by PG; it seemed weird to break compatibility just for the sake of config uniformity. I actually might just remove all the env vars. I don't like them, and ops doesn't plan to use them.

server/controllers/visit.js
```diff
@@ -0,0 +1,52 @@
...
+'use strict';
+
+var Boom = require('boom');
+
+var log = require('../logger')('server.controllers.visit');
+var visit = require('../models/visit');
+
+var visitController = {
+  get: function (request, reply) {
+    var userId = request.auth.credentials;
```
6a68: would be nice to actually store this on the controller

server/controllers/visits.js
```diff
@@ -0,0 +1,72 @@
...
+    var o = {
+      userId: userId,
+      visitId: visitId,
+      url: p.url,
+      urlHash: urlHash,
+      title: p.title,
+      visitedAt: p.visitedAt
+    };
+    queue.createVisit(o);
+    if (config.get('embedly.enabled')) {
```
6a68: this should be "if embedly is enabled AND if the user_page with that urlHash doesnt exist, then send it to the scraper"

server/controllers/visits.js
```diff
@@ -0,0 +1,72 @@
...
+var Boom = require('boom');
+var uuid = require('uuid');
+
+var config = require('../config');
+var log = require('../logger')('server.controllers.visits');
+var queue = require('../work-queue/queue');
+var visits = require('../models/visits');
+
+// TODO: add additional metadata fields to the visit datatype
+// TODO: normalize URLs
```
6a68: crap. definitely need to normalize URLs. the other TODO is done, though :-)

server/controllers/visits.js
```diff
@@ -0,0 +1,72 @@
...
+    var urlHash = crypto.createHash('sha1').update(p.url).digest('hex').toString();
+    var o = {
+      userId: userId,
+      visitId: visitId,
+      url: p.url,
+      urlHash: urlHash,
+      title: p.title,
+      visitedAt: p.visitedAt
+    };
+    queue.createVisit(o);
```
6a68: it would be nice for the job to check that the visit doesn't exist before attempting to create it, or just handle the "error, record already exists" error cleanly.

server/db/elasticsearch.js
```diff
@@ -0,0 +1,40 @@
...
+  timeout: config.get('db.elasticsearch.queryTimeout'),
+  // query elasticsearch; returns a promise
+  //
+  // queryType := es client API method, for example, 'create'
+  // params := es query DSL object
+  query: function query(queryType, params) {
+    var esClient = new es.Client(esParamsFactory());
+    // force promises to eventually resolve
+    // NOTE: I checked, elasticsearch uses bluebird, which supports Promise.timeout ^_^
+    return esClient[queryType](params).timeout(elasticsearch.timeout, 'elasticsearch timed out');
```
6a68: is it possible to convert the bluebird promise into a Q promise, so that we aren't worrying about API incompatibilities?

server/db/postgres.js
```diff
@@ -0,0 +1,79 @@
...
+      Object.keys(row).forEach(function(k) {
+        output[camelize(k)] = row[k];
+      });
+      outputRows.push(output);
+    });
+    return outputRows;
+  },
+  // query postgres; returns a promise
+  //
+  // query := a prepared query string
```
6a68: I guess these docs are nice, but it seems weird to only insert them in one spot.

server/embedly.js
```diff
@@ -0,0 +1,109 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+// Facade that abstracts the embed.ly endpoint.
+// TODO use Joi to validate the url and ensure key is defined
```
6a68: This.

Clarifying the contract between layers in the app makes it trivial to distribute the layers later. Having some type of schema validation also lets us split the problem set. Part of this contract should include not stupidly bubbling up errors all the way to the top level; DB errors should be handled at the DB, not returned to a model, which might return it upwards to a controller. Ugz.

server/embedly.js
```diff
@@ -0,0 +1,109 @@
...
+var log = logFactory('server.services.embedly');
+var npmLog = logFactory('server.services.embedly.vendor');
+
+module.exports = {
+  extract: function extract(url, cb) {
+    log.debug('embedly.extract called');
+
+    if (!isEnabled) {
+      return cb('Embedly is currently disabled. Set `server.embedly.enabled` to enable.');
+    }
```
6a68: Probably should also include a helpful error if the apiKey is undefined, since that seems like an easy thing to forget.

server/models/user-page.js
```diff
@@ -0,0 +1,116 @@
...
+var elasticsearch = require('../db/elasticsearch');
+var log = require('../logger')('server.models.user-page');
+
+var _verbose = function() {
+  var logline = [].join.call(arguments, ', ');
+  log.verbose(logline);
+};
+
+var userPage = {
+  _onFulfilled: function _onFulfilled(msg, callback, results) {
```
6a68: Clearly some duplication across all of the models. Maybe a base class? Or, maybe a shared mixin, favoring composition over inheritance.

server/models/user-page.js
```diff
@@ -0,0 +1,116 @@
...
+    _verbose(name + ' called', userId, url);
+    var lazyCreateParams = [uuid.v4(), userId, url, urlHash, title];
+    var lazyCreateUserPageQuery = 'WITH new_page AS (  ' +
+      '  INSERT INTO user_pages (id, user_id, url, raw_url, url_hash, title) ' +
+      '  SELECT $1, $2, $3, $3, $4, $5 ' +
+      '  WHERE NOT EXISTS (SELECT id FROM user_pages WHERE user_id = $2 AND url_hash = $4) ' +
+      '  RETURNING id ' +
+      ') SELECT id FROM new_page ' +
+      'UNION SELECT id FROM user_pages WHERE user_id = $2 AND url_hash = $4';
+    var addExtractedDataQuery = 'UPDATE user_pages SET ( ' +
```
6a68: It might be nice to have a really simple class that represents the fields and types in a table, to reduce this kind of silly verbosity...oh wait, I'm describing an ORM. It might be nice to use an ORM.

server/models/user-page.js
```diff
@@ -0,0 +1,116 @@
...
+    _verbose('about to issue lazy user page creation query');
+    var userPageId;
+    postgres.query(lazyCreateUserPageQuery, lazyCreateParams)
+      .fail(userPage._onRejected.bind(userPage, name + ' failed early', cb))
+      .then(function(result) {
+        // we just got the page_id; push it onto the end of the params array
+        _verbose('the lazy create result is ' + JSON.stringify(result));
+        _verbose('the userPageId is ' + result.id);
+        userPageId = result.id;
+        addDataParams.push(result.id);
```
6a68: I really dislike how clever this is. Maybe just define the params inside the callback.

server/models/user-page.js
```diff
@@ -0,0 +1,116 @@
...
+        var esQuery = {
+          index: 'chronicle',
+          type: 'userPages',
+          id: userPageId,
+          body: data
+        };
+
+        _verbose('the elasticsearch query is ' + JSON.stringify(esQuery));
+        // if the page doesn't exist yet, this task creates it. visit.create gives up if
+        // the page exists, preventing the scraper fields from being deleted.
```
6a68: pretty sure this comment is a lie

server/models/visit.js
```diff
@@ -0,0 +1,157 @@
...
+var visit = {
+  _onFulfilled: function _onFulfilled(msg, callback, results) {
+    _verbose(msg);
+    callback(null, results);
+  },
+  _onRejected: function _onRejected(msg, callback, err) {
+    log.warn(msg);
+    callback(err);
+  },
+  _transform: function _transform(results) {
```
6a68: I'm so pissed that we have this crap in the model. This belongs in a view object. One way to connect them: set the fields on the model, then fire an updated event so that the view can update itself. A duller but more direct way: controller passes model into view then calls view.render, or something, orchestrating it explicitly.

server/models/visit.js
```diff
@@ -0,0 +1,157 @@
...
+    var query = 'SELECT visits.id as visit_id, visits.user_id as user_id, * ' +
+    'FROM visits LEFT JOIN user_pages ON visits.user_page_id = user_pages.id ' +
+    'WHERE visits.id = $1 AND visits.user_id = $2';
+    var params = [visitId, userId];
+    postgres.query(query, params)
+      .then(function(results) {
+        // return a promise that resolves to the transformed results
+        return q(visit._transform(results));
+      })
+      .done(visit._onFulfilled.bind(visit, name + ' succeeded', cb),
```
6a68: /me holds nose. Can't we replace this with some nice sugary thing?

server/models/visit.js
```diff
@@ -0,0 +1,157 @@
...
+            title: title,
+            visitedAt: visitedAt
+          }
+        };
+        // if the scraper elasticsearch task happens first, this one will abort, to avoid
+        // overwriting all the scraped data with nothing
+        return elasticsearch.query('create', esQuery)
+          .then(null, function(err) {
+            // TODO if the error is due to the thing already existing, it's not a problem...
+            // swallowing errors for now, until I look up how to catch that specific error
```
6a68: this is a bad place to swallow all errors. fix.

server/models/visit.js
```diff
@@ -0,0 +1,157 @@
...
+    // delete the visit if it exists.
+    // delete the visit's userPage if no other visits have that page.
+    var name = 'models.visit.delete';
+    var userPageId;
+    postgres.query('DELETE FROM visits WHERE user_id = $1 AND id = $2 RETURNING user_page_id', [userId, visitId])
+      .fail(visit._onRejected.bind(visit, name + ' failed to delete visit', cb))
+      .then(function(result) {
+        _verbose('delete returning result gives us: ' + JSON.stringify(result));
+        userPageId = result.userPageId;
+        return postgres.query('SELECT count(*) FROM visits WHERE user_id = $1 AND user_page_id = $2',
```
6a68: does postgres remove entries from indexes synchronously or async? if async, this query might fail to detect userPages with no remaining visits; we should eventually add a job to look for orphaned userPages and delete them.

server/models/visit.js
```diff
@@ -0,0 +1,157 @@
...
+        // assuming count is int
+        if (count > 0) {
+          // nothing more to do here; call _onFulfilled.
+          // todo: does this ensure the promise chain ends?
+          visit._onFulfilled(name + ' succeeded', cb, null);
+        } else {
+          _verbose('no other visits exist for user page ' + userPageId + ': deleting it');
+          var deleteQuery = 'DELETE FROM user_pages WHERE user_id = $1 AND id = $2';
+          return postgres.query(deleteQuery, [userId, userPageId])
+            .then(elasticsearch.query('delete', {index: 'chronicle', type: 'userPages', id: userPageId}));
```
6a68: nesting promise chains is too clever. find another way

server/models/visits.js
```diff
@@ -0,0 +1,115 @@
...
+var visits = {
+  _onFulfilled: function _onFulfilled(msg, callback, results) {
+    _verbose(msg);
+    callback(null, results);
+  },
+  _onRejected: function _onRejected(msg, callback, err) {
+    log.warn(msg);
+    callback(err);
+  },
+  _transform: function _transform(results) {
```
6a68: ughhhhhhhh

server/models/visits.js
```diff
@@ -0,0 +1,115 @@
...
+    var name = 'models.visits.search';
+    _verbose(name + ' invoked', userId, count);
+    var esQuery = {
+      index: 'chronicle',
+      type: 'userPages',
+      size: count,
+      body: {
+        query: {
+          match: {
+            extractedContent: searchTerm
```
6a68: do better

server/work-queue/queue.js
```diff
@@ -97,6 +97,10 @@ module.exports = {
   createVisit: function(o) {
```
6a68: pretty sure there's supposed to be a callback here. look into it.

bin/create_db.sh
```diff
@@ -19,22 +19,66 @@ psql -c "CREATE DATABASE chronicle ENCODING 'UTF-8' LC_COLLATE = 'en_US.UTF-8' L
...
+);" -d chronicle -U chronicle
```
pdehaan: Maybe I'm missing it, but should we just dump the entire result from Embedly into the Postgres database in case we need any other fields in the future (ie: related items, error messages, etc)

bin/create_db.sh
```diff
@@ -19,22 +19,66 @@ psql -c "CREATE DATABASE chronicle ENCODING 'UTF-8' LC_COLLATE = 'en_US.UTF-8' L
...
+  extracted_media_duration INTEGER,
+  extracted_media_height INTEGER,
+  extracted_media_html TEXT,
+  extracted_media_type TEXT,
+  extracted_media_width INTEGER,
+  extracted_provider_display TEXT,
+  extracted_provider_name TEXT,
+  extracted_provider_url VARCHAR(2048),
+  extracted_published TIMESTAMPTZ(3),
+  extracted_safe BOOLEAN,
```
pdehaan: When `safe` is false, we also get two other fields from Embed.ly which we may need to store: http://embed.ly/docs/api/extract/features/safe

Also (from same doc):
> Once a url is flagged as malicious, you must respect the `cache_age` attribute. When `cache_age` expires, you must not show the message again until you have reevaluated that url with Embedly. This helps to mitigate false positives and ensure compliance with the terms of Google’s Safe Browsing API.

... Which I guess is OK since we're storing the `cache_age` above. We may just need a bit more logic around whether we need to priority rescrape a page to see if it is still on the blacklist.

server/db/elasticsearch.js
pdehaan: Is there a reason we use Q instead of Bluebird?

server/embedly.js
```diff
@@ -0,0 +1,109 @@
...
+var logFactory = require('./logger');
+var log = logFactory('server.services.embedly');
+var npmLog = logFactory('server.services.embedly.vendor');
+
+module.exports = {
+  extract: function extract(url, cb) {
+    log.debug('embedly.extract called');
+
+    if (!isEnabled) {
+      return cb('Embedly is currently disabled. Set `server.embedly.enabled` to enable.');
```
pdehaan: `server.embedly.enabled` or just `embedly.enabled`?

server/embedly.js
```diff
@@ -0,0 +1,109 @@
...
+          out.extractedPublished = new Date(publicationDateMillis).toJSON();
+        }
+
+        // not in the response, but we want extracted_at to be the current time
+        out.extractedAt = new Date().toJSON();
+
+        if (d.authors && d.authors.length) {
+          out.extractedAuthorName = d.authors[0].name;
+          out.extractedAuthorUrl = d.authors[0].url;
+        }
```
pdehaan: Why this? Why not just store all the authors?

bin/create_db.sh
6a68: Good catch! I'll make updated_at mandatory everywhere.

The reason for the change, btw, is that elasticsearch's [optimistic concurrency](http://www.elasticsearch.org/guide/en/elasticsearch/guide/current/optimistic-concurrency-control.html) model relies on queries including an optional version identifier; the queries will bail if the version is wrong. Since postgres is (for now) our primary datastore, we want to use the postgres `updated_at` timestamp as our ordered, sortable version identifier. But that's a todo, it's not implemented yet - either in ES or in our ES queries. So...maybe I'll just ensure updated_at is always there, for now, in all the tables.

bin/create_db.sh
6a68: I considered that, but these are the fields Nick settled on in v0, and I'd rather not keep extra stuff around for some possible future use. If we decide we're missing some vital bit of data, we can test that theory by reimporting recent browser history with the extra fields added, and see if the product's better or not ^_^

config/test-urls.js
```diff
@@ -0,0 +1,101 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+// some of the alexa US top 50 sites and subpages
+module.exports = [
+  { url: 'https://www.google.com/?gws_rd=ssl',
```
6a68: I don't explicitly call this out anywhere, but one of my goals with this file is to make it really easy to generate specific training sets, like clusters of pages around a search, or big sets of social or media-focused pages. I opted for `.js` over `.json` because I hate how you can never cleanly insert comments in a JSON file; comments are nice.

grunttasks/build.js
6a68: One of my primary use cases is: something's busted, let's insert a `debugger` statement and step through it with node-debugger. But the linter complains every time, and every time I disable it temporarily in the gruntfile. Might as well optimize for actual _development_ on the server.

package.json
```diff
@@ -20,6 +21,7 @@
     "pg": "4.1.1",
     "q": "1.1.2",
     "underscore": "1.7.0",
+    "underscore.string": "^3.0.1",
```
6a68: derp, I'll get rid of that `^`

server/README.md
```diff
@@ -0,0 +1,94 @@
+## Working with the server code
+
+### Code "structure"
```
6a68: LIES!

server/controllers/visits.js
```diff
@@ -0,0 +1,72 @@
...
+
+'use strict';
+
+var crypto = require('crypto');
+var Boom = require('boom');
+var uuid = require('uuid');
+
+var config = require('../config');
+var log = require('../logger')('server.controllers.visits');
+var queue = require('../work-queue/queue');
```
6a68: I'd really prefer to publish events for the queue, rather than directly call the queue. I'd like to fix this when we add a global event dispatcher to the app, hopefully next sprint.

server/embedly.js
```diff
@@ -0,0 +1,109 @@
...
+          out.extractedImageCaption = d.images[0].caption;
+          // another array of rgb values
+          if (d.images.colors) {
+            out.extractedImageColor = d.images[0].colors[0].color;
+          }
+        }
+
+        if (d.media && d.media.type) {
+          out.extractedMediaType = d.media.type;
+          out.extractedMediaHtml = d.media.html;
```
pdehaan: I'll have to play around with this a bit more locally and scrape more pages, but not sure if we always get `html` back. http://embed.ly/docs/api/extract/endpoints/1/extract#media

For "photo" types, I think we only get back a URL (and no `duration`):
```js
  "media": {
    "height": 330,
    "type": "photo",
    "url": "https://crazycrawfish.files.wordpress.com/2013/12/121613_1703_therumorsof1.jpg?w=594",
    "width": 594
  },
```

server/models/user-page.js
```diff
@@ -0,0 +1,116 @@
...
+  var logline = [].join.call(arguments, ', ');
+  log.verbose(logline);
+};
+
+var userPage = {
+  _onFulfilled: function _onFulfilled(msg, callback, results) {
+    _verbose(msg);
+    callback(null, results);
+  },
+  _onRejected: function _onRejected(msg, callback, err) {
```
pdehaan: Maybe I'm not seeing a bigger picture, but the `err` argument as the last item seems hawkward to me. But maybe not something in our control.

server/models/user.js
```diff
@@ -0,0 +1,69 @@
...
+  _onFulfilled: function _onFulfilled(msg, callback, results) {
+    _verbose(msg);
+    _verbose('onfulfilled results are: ' + JSON.stringify(results));
+    callback(null, results);
+  },
+  _onRejected: function _onRejected(msg, callback, err) {
+    log.warn(msg);
+    callback(err);
+  },
+  create: function(userId, email, oauthToken, cb) {
```
pdehaan: minor point, but some mild inconsistencies here w/ `callback` vs `cb` argument names (probably a more global issue).

server/db/elasticsearch.js
6a68: Good question! I don't have a great answer; Q came first, I know it a little better. I'm pretty sure there is API parity at this point, but not totally sure (hence my slightly paranoid code comment above). I think bluebird bills itself as performant, but we're using promises for slow stuff like DB calls, so I dunno if that's terribly important here; we'll use pubsub for most communication within the app, rather than promises, as time goes on. I'm happy to hear your thoughts on why bluebird might be better :-)

server/embedly.js
```diff
@@ -0,0 +1,109 @@
...
+      api.extract({url:url}, function(err, data) {
+        log.verbose('embedly.extract callback for url ' + url + ' returned data ' + JSON.stringify(data));
+        if (err) { return cb(err); }
+        var d = data && data[0];
+        if (!d) { return cb(new Error('embedly response was empty for url ' + url)); }
+
+        // for right now, let's just return the keys we want in user_pages
+        var out = {};
+
+        // these keys are top-level and match the format in the database
```
6a68: it'd be nice to split all these stanzas into tiny helper functions, would make them easier to test / isolate bugs

server/models/user.js
```diff
@@ -0,0 +1,69 @@
...
+    _verbose(name + ' called', userId, email, oauthToken);
+    var query = 'INSERT INTO users (user_id, email, oauth_token, created_at) ' +
+                'VALUES ($1, $2, $3, $4)';
+    var params = [userId, email, oauthToken, new Date().toJSON()];
+    postgres.query(query, params)
+      .done(user._onFulfilled.bind(user, name + ' succeeded', cb),
+            user._onRejected.bind(user, name + ' failed', cb));
+  },
+  // TODO TODO use this! :-)
+  exists: function(userId, cb) {
```
6a68: Do we really want to use this instead of user.get? I don't know if we need both functions, since the only user.get is inside the session initialization code.

server/models/visit.js
```diff
@@ -0,0 +1,157 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+var q = require('q');
```
pdehaan: Minor thing, but I notice it is `var q` here and `var Q` in other files (see _server/models/user.js_).

server/models/visits.js
```diff
@@ -0,0 +1,115 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+var q = require('q');
```
pdehaan: See [previous comment](https://github.com/mozilla/chronicle/pull/224/files#r23794554) re: changing this to
```js
var Q = require('q');
```

server/models/visit.js
6a68: it's fine for now. too many layers of indirection is an obstacle to progress.

server/models/visit.js
```diff
@@ -0,0 +1,157 @@
...
+      ') SELECT id FROM new_page ' +
+      'UNION SELECT id FROM user_pages WHERE user_id = $2 AND url_hash = $4';
+    var lazyCreateParams = [uuid.v4(), userId, url, urlHash, title, new Date().toJSON()];
+    var createVisitQuery = 'INSERT INTO visits ' +
+      '(id, user_id, user_page_id, visited_at, updated_at) ' +
+      'VALUES ($1, $2, $3, $4, $4)';
+    var userPageId;
+
+    // try to insert into pg; handle pg errors; insert into es; then we're done.
+    // don't pass results to fulfillment callback, because they aren't used by the caller.
```
6a68: actually, if create returns the record, then the worker can verify that the returned results look sane (or not). something to consider.

server/models/visit.js
```diff
@@ -0,0 +1,157 @@
...
+    postgres.query(lazyCreateUserPageQuery, lazyCreateParams)
+      .fail(visit._onRejected.bind(visit, name + ' failed', cb))
+      .then(function(results) {
+        // TODO will userPageId be correct? find out :-P
```
6a68: remove this TODO

server/models/visit.js
```diff
@@ -0,0 +1,157 @@
...
+        userPageId = results && results.id;
+        var visitParams = [visitId, userId, userPageId, visitedAt];
+        // return the new query's promise to avoid nesting promise chains
```
6a68: not a useful comment, remove

server/models/visits.js
```diff
@@ -0,0 +1,115 @@
...
+
+var config = require('../config');
+var postgres = require('../db/postgres');
+var elasticsearch = require('../db/elasticsearch');
+var log = require('../logger')('server.models.visits');
+var visit = require('./visit');
+
+var _verbose = function() {
```
6a68: maybe a base model class with this and other copypasta'd functions like _onFulfilled could temporarily work in place of the inheritance v. composition mental debate

server/work-queue/jobs/extract-page.js
```diff
@@ -0,0 +1,36 @@
...
+
+module.exports = {
+  // o is an object with keys { userId, url, urlHash, title }
+  perform: function(o, cb) {
+    log.verbose('job created with params ' + JSON.stringify(o));
+    embedly.extract(o.url, function(err, data) {
+      if (err) {
+        log.warn('job failed, throwing to trigger a retry');
+        // we should retry on failure. leave that to the queue.
+        throw err;
```
6a68: `cb(err)`

server/work-queue/jobs/extract-page.js
```diff
@@ -0,0 +1,36 @@
...
+
+var log = require('../../logger')('server.work-queue.jobs.extract-url');
+var userPage = require('../../models/user-page');
+var embedly = require('../../embedly');
+
+module.exports = {
+  // o is an object with keys { userId, url, urlHash, title }
+  perform: function(o, cb) {
+    log.verbose('job created with params ' + JSON.stringify(o));
+    embedly.extract(o.url, function(err, data) {
```
6a68: maybe we can group embedly.js, postgres.js, elasticsearch.js, and queue.js under "things which require network IO", and create some kind of unified interface/contract for all of them. Not just promises or callbacks, but an API and some standard object properties. (Wouldn't it be nice for __heartbeat__ to just do something like: "ask each of the Service objects if they are able to connect"?

server/work-queue/jobs/extract-page.js
```diff
@@ -0,0 +1,36 @@
...
+        // we should retry on failure. leave that to the queue.
+        throw err;
+      } else {
+        log.verbose('extract-page job succeeded for user ' + o.userId + ', url ' + o.url + ': ' + JSON.stringify(data));
+        // the visit creation job has probably created a record in the user_page table.
+        // if not, update will lazily create it.
+        userPage.update(o.userId, o.url, o.urlHash, o.title, data, function (err) {
+          if (err) {
+            log.warn('userPage.update failed: ' + err);
+            throw err;
```
6a68: ditto

server/routes/index.js
```diff
@@ -4,4 +4,11 @@
 
 'use strict';
 
-module.exports = [].concat(require('./base'), require('./auth'), require('./search'), require('./visits'));
+// TODO just sniff out all the files under /routes?
```
pdehaan: Possibly [**node-require-directory**](https://www.npmjs.com/package/node-require-directory) or [**require-directory**](https://www.npmjs.com/package/require-directory).

server/embedly.js
6a68: doh! great catch, missed that one

server/embedly.js
6a68: we're storing one of everything for now; that's the most I could see us using in the current UI

server/models/user-page.js
6a68: totally agree, but this whole `.done(foomodel._onFulfilled.bind(null ... ), foomodel._onRejected.bind(null...)` thing needs work anyway; I predict it vanishes sooner than later

server/routes/index.js
6a68: thanks for the suggestions! `node-require-directory` shows me generated coffeescript, and `node-require` seems like massive overkill. I suspect there's a one-liner lurking here that'd be sufficient for our current case; care to take a swing?

server/db/elasticsearch.js
pdehaan: Oh, I don't care one way or the other. Personally, I use [then/promise](https://www.npmjs.com/package/promise), but no real reason why I chose that over anything else. I think we use bluebird for [fxa-content-server](https://github.com/mozilla/fxa-content-server/blob/85e784679d149f8c75e445bf050617e449711130/package.json#L26) so I figured it was probably battle tested. We can use whatever you think is best, I was just thinking if you were concerned about bluebird+Q compatibility, it may be easiest just to stick with whatever one was being returned by the Elasticsearch module.

server/work-queue/jobs/extract-page.js
```diff
@@ -0,0 +1,36 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+var log = require('../../logger')('server.work-queue.jobs.extract-url');
```
pdehaan: `extract-url` or `extract-page` (considering the filename)?

server/controllers/visits.js
6a68: fixed

server/controllers/visits.js
6a68: removed

server/routes/index.js
pdehaan: I can try taking a look. The `node-require-directory` has some pretty solid logic, although I'm having issues with getting it to include the current directory (so we may need to put the routes into a subdirectory &mdash; `server/routes/routes/base.js` instead of `server/routes/base.js`). Then we could use this one liner in `server/routes/index.js`:

```js
var requireDir = require('node-require-directory');

module.exports = requireDir('routes');
```

I tried all sorts of variants of `requireDir('.')` and `requireDir(__dirname)` but neither worked, as I'm guessing it's hitting some weird recursive thing with the index.js trying to include itself, which is why I'm thinking it needs to have all the route files in a subdirectory.

6a68: I was thinking something along the lines of
```js
// inside index.js
var fs = require('fs');
var routes = [];
fs.readdirSync(__dirname).forEach(function(file) {
  if (file != 'index.js') { routes.concat(require('./' + file)); }); }
});
module.exports = routes;
```
What do you think? Am I missing anything?

server/work-queue/jobs/extract-page.js
6a68: doh!

bin/create_test_data.js
```diff
@@ -8,114 +8,82 @@
...
-        cb && cb(err);
-      }
+    return cb && cb(err);
+  });
+}
 
-      if (err) {
-        return donezo('failed to connect to database', err);
+function createTestData(cb) {
+  var userId = config.get('testUser.id');
```
pdehaan: ESLint says that we aren't using `userId` anywhere. Not sure if we want to remove this, or possibly change the `generateTestRequest()` `auth.credentials` value below on line 42 to use the value (instead of requerying the config for each loop iteration.

server/db/elasticsearch.js
```diff
@@ -0,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+var es = require('elasticsearch');
+var Q = require('q');
```
pdehaan: I don't think we're using `Q` in this file (unless you're planning on doing that Bluebird->Q promise converter).

server/db/elasticsearch.js
```diff
@@ -0,0 +1,40 @@
...
+
+var config = require('../config');
+var log = require('../logger')('server.db.elasticsearch');
```
pdehaan: Also, I don't think we're using `log()` anywhere [yet].

server/db/postgres.js
```diff
@@ -0,0 +1,79 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+var pg = require('pg');
+var Q = require('q');
+var camelize = require('underscore.string').camelize;
+var underscored = require('underscore.string').underscored;
```
pdehaan: Not sure that we're using `underscored` here.

server/db/postgres.js
```diff
@@ -0,0 +1,79 @@
...
+
+// primitive DB object
+var postgres = {
+  timeout: config.get('db.postgres.queryTimeout'),
+  // postgres is case-insensitive, so we store keys as underscored there,
+  // and transform back to camelCase when returning results here.
+  // individual queries must be sure to use the underscored versions.
+  camelize: function(rows) {
+    var outputRows = [];
+    var output, k;
```
pdehaan: Don't think we need `k` here, as we're clobbering it as a function arg in line 35.

bin/create_db.sh
6a68: Hmm. The addon won't send down a visit until the user has actually visited the page. I'm guessing their browser will tell them this is a suspected bad-news page; if they choose to click through the warning and visit the page, I think it makes sense for the visit to show up in their history (as a poorly scraped visit with a lot of missing fields).

bin/create_db.sh
6a68: fixed

package.json
6a68: fixed

server/README.md
6a68: removed

server/controllers/visit.js
6a68: do it later

server/controllers/visits.js
6a68: later

server/controllers/visits.js
6a68: added

server/models/user-page.js
```diff
@@ -0,0 +1,116 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+// we need a user-pages abstraction so that async scraper jobs
+// can separately handle creating the visit and creating the
+// user page. someday this might also be used by the client-facing API layer.
+
+var underscored = require('underscore.string').underscored;
```
pdehaan: `underscored` is defined but never used.

server/models/user.js
```diff
@@ -0,0 +1,69 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+var Q = require('q');
```
pdehaan: `Q` is defined but never used.

server/models/user.js
```diff
@@ -0,0 +1,69 @@
...
+
+var config = require('../config');
```
pdehaan: `config` is defined but never used.

server/db/elasticsearch.js
6a68: good point! let's file that under "things to do later but soonish maybe"


(it will be nice to stop putting off perfectly reasonable decisions someday)

server/db/postgres.js
6a68: fixed.js!

server/db/postgres.js
6a68: yissssss, fixed

server/db/postgres.js
6a68: meh

server/embedly.js
6a68: I feex.

server/embedly.js
6a68: added

server/embedly.js
6a68: meh

server/embedly.js
6a68: as long as we have a media object (i.e. we don't try to access a property of an undefined thing and generate a `TypeError`), it's fine to have some nulls in there. `publisher` and `authors` rarely seem to succeed, for instance.

server/models/user-page.js
6a68: good catch! fixed

server/models/user-page.js
6a68: meh

server/models/visits.js
```diff
@@ -0,0 +1,115 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+var q = require('q');
+
+var config = require('../config');
```
pdehaan: `config` is defined but never used.

server/models/visit.js
```diff
@@ -0,0 +1,157 @@
...
+  },
+  _onRejected: function _onRejected(msg, callback, err) {
+    log.warn(msg);
+    callback(err);
+  },
+  _transform: function _transform(results) {
+    // do a little transformin' and this might barf, so do it before `done`
+
+    // if nothing was found, continue
+    if (!results) { return; }
```
pdehaan: ESLint is complaining about "Expected no return value" on line 45. I think it's because up here on line 32 we're just calling `return` with no args or return value, whereas on line 45 we're returning an Object. Not sure if we should return an empty object here, or just ignore whatever ESLint think's it's validating on the [`consistent-return`](http://eslint.org/docs/rules/consistent-return.html) rule.

> _One of the confusing aspects of JavaScript is that any function may or may not return a value at any point in time. When a function exits without any `return` statement executing, the function returns `undefined`. Similarly, calling `return` without specifying any value will cause the function to return `undefined`. Only when `return` is called with a value is there a change in the function's return value. ..._

All moot if your calling code checks for a valid return value vs `undefined` though. Feel free to stop reading my nonsense at any time...

server/models/user-page.js
6a68: http://media.giphy.com/media/pCDPxNX3j3XJS/giphy.gif

server/models/user-page.js
6a68: m e h . i o

server/models/user-page.js
6a68: removed

server/models/visits.js
```diff
@@ -0,0 +1,115 @@
...
+    log.warn(msg);
+    callback(err);
+  },
+  _transform: function _transform(results) {
+    if (!results || !results.length) { return; }
+    var transformed = [];
+    results.forEach(function(result) {
+      transformed.push(visit._transform(result));
+    });
+    return transformed;
```
pdehaan: Same silly comment as https://github.com/mozilla/chronicle/pull/224/files#r23810436
I think ESLint is warning against returning an Object on line 35 vs a naked `return` on line 30 (which would translate to `return undefined`).

server/models/user.js
6a68: true, I still think _on* are going to be gone real soon

server/models/user.js
6a68: let's be explicit about exists vs get, it's done in several models by now.

server/models/visit.js
6a68: word. fixxxxx

6a68: oh, I remember what it was. The friggin linter complains about `Q(...)`, thinking it's a constructor. I'll just leave it for the moment, probably downcase the other `Q`s at some point

server/routes/auth.js
```diff
@@ -6,11 +6,13 @@
 
 var log = require('../logger')('server.routes.auth');
```
pdehaan: `log` is defined but never used.

server/models/visit.js
6a68: removed

server/models/visit.js
6a68: removed

server/routes/auth.js
```diff
@@ -6,11 +6,13 @@
...
 var config = require('../config');
```
pdehaan: `config` is defined but never used.

server/models/visit.js
6a68: meh. let's just throw all errors for now.

server/routes/auth.js
```diff
@@ -57,26 +40,12 @@ module.exports = [{
   method: ['GET', 'POST'],
   path: '/auth/complete',
   config: {
+    handler: authController.complete,
     auth: {
       strategy: 'oauth',
       mode: 'try'
     },
```
pdehaan: Bonus trailing comma here. :shrug:

server/models/visit.js
6a68: currently synchronous, but we'll change it for performance at some point to `create index concurrently`. let's worry about this later.

server/models/visit.js
6a68: mehhh

server/routes/base.js
```diff
@@ -22,10 +24,6 @@ module.exports = [{
         redirectTo: false // don't redirect users who don't have a session
       }
     },
```
pdehaan: Bonus trailing comma here. :shrug_more:

server/models/visits.js
6a68: yeah, same issue as above, see line 49

server/models/visits.js
6a68: follow-on commit I think

server/work-queue/jobs/extract-page.js
6a68: fix'd

server/work-queue/jobs/extract-page.js
6a68: good idea, todo for later

server/work-queue/jobs/extract-page.js
6a68: :heavy_check_mark:

server/work-queue/jobs/extract-page.js
6a68: :heavy_check_mark:

server/work-queue/queue.js
6a68: :heavy_check_mark: we want this to be fire-and-forget; the callback we care about connects the queue to the job

server/routes/index.js
pdehaan: Yeah, I imagine that should work as long as we don't need recursion and only ever have *.js files in this directory.

server/routes/search.js
```diff
@@ -7,31 +7,22 @@
 var Joi = require('joi');
 var Boom = require('boom');
```
pdehaan: `Boom` is defined but never used.

server/routes/search.js
```diff
@@ -7,31 +7,22 @@
...
 
+var searchController = require('../controllers/search');
 var log = require('../logger')('server.routes.search');
```
pdehaan: `log` is defined but never used.

server/routes/visit.js
```diff
@@ -0,0 +1,56 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+var Joi = require('joi');
+
+var log = require('../logger')('server.routes.visit');
```
pdehaan: `log` is defined but never used.

server/routes/visits.js
```diff
@@ -6,76 +6,28 @@
 
 var crypto = require('crypto');
```
pdehaan: `crypto` is defined but never used.

server/routes/visits.js
```diff
@@ -6,76 +6,28 @@
...
-var Boom = require('boom');
-var uuid = require('uuid');
-var log = require('../logger')('server.routes.visits');
-var models = require('../models');
-var visit = models.visit;
-var visits = models.visits;
 
-// TODO: add additional metadata fields to the visit datatype
-// TODO: normalize URLs
+var log = require('../logger')('server.routes.visits');
```
pdehaan: `log` is defined but never used.

server/routes/index.js
6a68: well, let's just do it, I'll add it to the branch

bin/create_test_data.js
6a68: yep! fixed

server/work-queue/queue.js
```diff
@@ -32,7 +32,7 @@ var multiWorkerOpts = {
   connection: connectionDetails,
   queues: ['chronicle'],
   minTaskProcessors: 1,
-  maxTaskProcessors: 10,
+  maxTaskProcessors: 30,
```
pdehaan: trailing comma. :shrug:

server/db/elasticsearch.js
6a68: great, left that coercion thing as a todo further down.

package.json
pdehaan: Not sure if you want to remove the `^` semver range operator here. I don't think it does any harm since we're probably locking the version number via npm-shrinkwrap.json anyways.

bin/create_db.sh
```diff
@@ -19,22 +19,66 @@ psql -c "CREATE DATABASE chronicle ENCODING 'UTF-8' LC_COLLATE = 'en_US.UTF-8' L
...
-  ON visits (fxaId, visitedAt, id);" -d chronicle -U chronicle
+psql -c "CREATE UNIQUE INDEX user_id_visited_at_id
+  ON visits (user_id, visited_at, id);" -d chronicle -U chronicle
+# used to check if a user_page should be deleted on visit delete
+psql -c "CREATE UNIQUE INDEX user_id_user_page_id_id
+  ON visits (user_id, user_page_id, id);" -d chronicle -U chronicle
 
+# lest we forget! clean out our elasticsearch index, too
+curl -XDELETE 'http://localhost:9200/chronicle/'
+curl -XPUT 'http://localhost:9200/chronicle/'
```
pdehaan: I *think* this works.

I installed Elasticsearch today via brew, which required me to install something called "Java". After starting Elasticsearch via `$ elasticsearch --config=/usr/local/opt/elasticsearch/config/elasticsearch.yml`, I got the following output  after running `$ ./bin/create_db.sh`:
```sh
$ ./create_db.sh

DROP DATABASE
REASSIGN OWNED
DROP ROLE
CREATE ROLE
CREATE DATABASE
GRANT
ALTER SCHEMA
CREATE TABLE
CREATE TABLE
CREATE INDEX
CREATE TABLE
CREATE INDEX
CREATE INDEX
{"error":"IndexMissingException[[chronicle] missing]","status":404}
{"acknowledged":true}
```

Running the same command again gives me this slightly different output:
```sh
$ ./create_db.sh

DROP DATABASE
REASSIGN OWNED
DROP ROLE
CREATE ROLE
CREATE DATABASE
GRANT
ALTER SCHEMA
CREATE TABLE
CREATE TABLE
CREATE INDEX
CREATE TABLE
CREATE INDEX
CREATE INDEX
{"acknowledged":true}
{"acknowledged":true}
```

SO, long story longer, it seems that the first time I run the create_db.sh script, it may give me an inert warning message about the not-yet-created "chronicle" index being missing.

pdehaan:
OK, I think I've BSed all the feedback I can think of.
Firstly, and most importantly, this is SUPER awesome! great job  @6a68!! :star::star::star::star::star:
Secondly, @nchapman pre-approved of everything you do, so feel free to merge this when you're happy.
Thirdly, if there are any issues/todos/comments below that you want me to convert into bugs, let me know and I'll be happy to copy-paste my way to success.

server/work-queue/queue.js
6a68: heh, I'm wearing you down! I'll fix.

6a68:
Thanks for taking the time to read all teh codez, @pdehaan!

I missed your last round of comments, so we'll have to handle those in a follow-on PR (sorry!). I've got a squashed and rebased branch which Imma push momentarily, closing this review PR for now.

PR #226, opened 2015-01-30
feat(server): add scraper and lotsa other stuff

### LINKED ISSUE #21: add scraper service (extract interesting data from URLs)
Thoughts

- new URLs should be added to scraper queue
- scraper output should be JSON structured & sent to elasticsearch, it'll also need to be transformed and inserted into MySQL
- scraper-worker can just poll the scraper's endpoint (or wait for a callback, whatever), then chuck it in elasticsearch when ready
- scraper might be third party, or we might own it and build it ourselves
- scraper will need to canonicalize URL, generate summary, find a suitable image/media blob, and possibly also generate keywords
- let's have a provider-agnostic interface/contract that lets us use the same API whether it's embedly, our own scraper, or some combination of the two

- also: be sure to insert canonical URL into `visits.url` and add visits fields for other new fields

Tasks

- [ ] Create schema for user pages to store scraped data
- [ ] Create scraper worker (embedly)
- [ ] Index scraped data
- [ ] Expose scraped data in visits API
- [ ] Expose scraped data in search API
- [ ] Show scraped data on the front-end in the visits index

6a68:
Fixes #21 

For PR discussion see: https://github.com/mozilla/chronicle/pull/224

Merging as a crazy gigantic monolithic commit

**VICTORY IS OURS**


6a68:
welp, travis is failing, I can't repro it locally, so I'm just going to "break" travis for the moment, and fix it in the morning. many :beers: to all and to all a good night

6a68:
wait, I was wrong - this time, travis passed: https://travis-ci.org/mozilla/chronicle/builds/48847688

so, um, yay everything

PR #232, opened 2015-01-30
WIP: send welcome email

pdehaan:
Questions:
- [**sendmail**](https://www.npmjs.com/package/sendmail) vs [**nodemailer**](https://www.npmjs.com/package/nodemailer) -- It looks like nodemailer has support for different transport options (such as Amazon SES via [**nodemailer-ses-transport**](https://github.com/andris9/nodemailer-ses-transport)), plugins (DKIM), HTML+plaintext...
- When should we use SES vs just SMTP? Currently just has a big `if (config.get('env') === 'local') {...}` block, but not sure if we should use SMTP when `env` is "test".
- There has to be a better way of doing templates than my hacky thing of using `_.template` (since we were already requiring underscore in package.json). Not sure if we should be using Handlebars or something else. I need to dig into what content/auth server is doing to see if there is a more official way.
- Should we have different queue tasks for welcome emails vs other emails, or just switch statements in the server/work-queue/jobs/send-email.js to say `if (o.type === 'welcome') {...}` versus just having specific jobs like "send-welcome-email.js".

### DIFF
--- a/config/local.json.example
+++ b/config/local.json.example
@@ -27,6 +27,15 @@
     "apiKey": "",
     "secretKey": ""
   },
+  "email": {
+    "fromEmail": "info@chronicle.firefox.com",
+    "smtp": {
+      "host": "localhost",
+      "port": 25,
+      "auth_user": "",
+      "auth_pass": ""
+    }
+  },
   "server": {
     "host": "127.0.0.1",
     "port": 8080,
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -9,12 +9,12 @@
       "dependencies": {
         "cryptiles": {
           "version": "2.0.4",
-          "from": "cryptiles@>=2.0.0 <3.0.0",
+          "from": "cryptiles@2.x.x",
           "resolved": "https://registry.npmjs.org/cryptiles/-/cryptiles-2.0.4.tgz"
         },
         "hoek": {
           "version": "2.11.0",
-          "from": "hoek@>=2.0.0 <3.0.0",
+          "from": "hoek@^2.2.x",
           "resolved": "https://registry.npmjs.org/hoek/-/hoek-2.11.0.tgz"
         }
       }
@@ -26,7 +26,7 @@
       "dependencies": {
         "hoek": {
           "version": "2.11.0",
-          "from": "hoek@>=2.0.0 <3.0.0",
+          "from": "hoek@2.x.x",
           "resolved": "https://registry.npmjs.org/hoek/-/hoek-2.11.0.tgz"
         }
       }
@@ -53,32 +53,32 @@
               "dependencies": {
                 "nomnom": {
                   "version": "1.8.1",
-                  "from": "nomnom@>=1.5.0",
+                  "from": "nomnom@>= 1.5.x",
                   "resolved": "https://registry.npmjs.org/nomnom/-/nomnom-1.8.1.tgz",
                   "dependencies": {
                     "underscore": {
                       "version": "1.6.0",
-                      "from": "underscore@>=1.6.0 <1.7.0",
+                      "from": "underscore@~1.6.0",
                       "resolved": "https://registry.npmjs.org/underscore/-/underscore-1.6.0.tgz"
                     },
                     "chalk": {
                       "version": "0.4.0",
-                      "from": "chalk@>=0.4.0 <0.5.0",
+                      "from": "chalk@~0.4.0",
                       "resolved": "https://registry.npmjs.org/chalk/-/chalk-0.4.0.tgz",
                       "dependencies": {
                         "has-color": {
                           "version": "0.1.7",
-                          "from": "has-color@>=0.1.0 <0.2.0",
+                          "from": "has-color@~0.1.0",
                           "resolved": "https://registry.npmjs.org/has-color/-/has-color-0.1.7.tgz"
                         },
                         "ansi-styles": {
                           "version": "1.0.0",
-                          "from": "ansi-styles@>=1.0.0 <1.1.0",
+                          "from": "ansi-styles@~1.0.0",
                           "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-1.0.0.tgz"
                         },
                         "strip-ansi": {
                           "version": "0.1.1",
-                          "from": "strip-ansi@>=0.1.0 <0.2.0",
+                          "from": "strip-ansi@~0.1.0",
                           "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-0.1.1.tgz"
                         }
                       }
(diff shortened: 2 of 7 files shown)


nchapman:
> When should we use SES vs just SMTP

This reminds me that @mostlygeek mentioned using SES's SMTP gateway to make it easier to switch between services. Would that make things easier or harder?

6a68:
About the questions in the description:
- sendmail vs nodemailer: I have no idea. Take a guess; we'll iterate as we actually get towards production.
- SES vs SMTP: I like Benson's idea of always using an SMTP gateway; that should hopefully make the code simpler
- Templating: I think _.template is overkill, just do string concatenation and move along
- Other queue tasks: I think it would be fine, for now, to just call it the `send-welcome-email` job. When we add other email tasks, we'll be able to decide the right level of generality / code-sharing.

6a68:
btw really nice work! this looks great

server/work-queue/jobs/send-email.js
```diff
@@ -0,0 +1,48 @@
...
+module.exports = {
+  // o is an object with keys { email }
+  perform: function(o, cb) {
+    var toEmail = o.email;
+    log.verbose('job created with params ' + JSON.stringify(o));
+
+    switch (config.get('env')) {
+      case 'local':
+        sendmail({
+          from: 'info@chronicle.firefox.com',
```
6a68: I'm pretty sure we will need to vary the 'from' address. Maybe that should be a sendmail config?

server/work-queue/jobs/send-email.js
```diff
@@ -0,0 +1,48 @@
...
+
+var contentTmpl = _.template('Welcome to Chronicle, <%= email %>!');
+
+module.exports = {
+  // o is an object with keys { email }
+  perform: function(o, cb) {
+    var toEmail = o.email;
+    log.verbose('job created with params ' + JSON.stringify(o));
+
+    switch (config.get('env')) {
```
6a68: Let's just stick the `local` env case inline here and remove all these other cases and remove the switch statement / env check completely. We'll get there when we get there. Travis / intern don't currently exercise the new user path, so we don't have to solve the `test` env right now.

server/work-queue/jobs/send-email.js
```diff
@@ -0,0 +1,48 @@
...
+      case 'local':
+        sendmail({
+          from: 'info@chronicle.firefox.com',
+          to: toEmail,
+          subject: 'Welcome to Chronicle',
+          content: contentTmpl({email: toEmail})
```
6a68: yeah, seriously, I'd just do `content: 'welcome to chronicle, ' + email + '!'`

server/work-queue/jobs/send-email.js
```diff
@@ -0,0 +1,48 @@
...
+        }, function (err, reply) {
+          if (err) {
+            log.error('Unable to send email: ' + err);
+            cb(err);
+            throw err;
```
6a68: I played with the queue a bit more yesterday; I think `cb(err)` tells node-resque the job failed, and we don't need to throw. Sorry I've been unclear on this point.

6a68:
@pdehaan looks great, man! feel free to ping me if you have followup questions, otherwise this is fine to land once it's updated.

6a68:
oh wait, I guess it needs to be rebased. Ping me if that goes sideways :-P

6a68:
After looking a bit more, I think nodemailer might be a better bet than sendmail, only because it has way more people using it, and it's been steadily maintained. (Maybe I'm wrong, and sendmail is done in some sense, because it has very few features? Feel free to dig further and make the final call here.)

pdehaan:
OK, attempt 2. Rip it apart.

TL;DR:
- added some configs to convict so we can customize per environment (and updated config/local.json.example -- although possibly foolishly since "local" doesn't use SES):
```js
email: {
  fromEmail: {
    format: 'email',
    default: 'info@chronicle.firefox.com'
  },
  ses: {
    accessKeyId: {
      format: String,
      default: 'your_key_here'
    },
    secretAccessKey: {
      format: String,
      default: 'your_key_here'
    }
  }
},
```
- changed from **sendmail** to **nodemailer**.
- using hypothetical AWS SES transport if `NODE_ENV` != "local".
- renamed to `send-welcome-email` and queue job name to `sendWelcomeEmail`.
- sends plaintext and equally ugly HTML welcome emails (thanks to the mighty power of the `<strong>` tag).

mostlygeek:
I would suggest skipping the SES api and just use SMTP. If you need a local one then we can run postfix (or whatever). 

When we authenticate to SES SMTP we will use a username/password. This be different from an AWS IAM access key/pass. So you should have a smtp user/password in convict. 

And for my sanity, please have a *flat* convict schema. 

Ben says no to this: 
````
email: {
  fromEmail: {
    format: 'email',
    default: 'info@chronicle.firefox.com'
  },
  ses: {
    accessKeyId: {
      format: String,
      default: 'your_key_here'
    },
    secretAccessKey: {
      format: String,
      default: 'your_key_here'
    }
  }
},
````

Having a deeply nested schema makes it impossible for us to merge default conf + env confs + secrets into a single JSON object during deployment. Every time a recursive deep merge is required that cute science doggy gets no dinner.

File: config/local.json.example [hunk 5]
```diff
@@ -27,6 +27,13 @@
     "apiKey": "",
     "secretKey": ""
   },
+  "email": {
+    "fromEmail": "info@chronicle.firefox.com",
```
6a68: Email sent with this from address never arrived in my gmail account. I don't know what a good solution is here, though, so we can just leave it.

6a68:
@pdehaan looks great, works for me, feel free to rebase + merge

6a68:
How you ask?
  1.  `git rebase -i HEAD~3` from this branch (because we want to squash the most recent 3 commits together)
  2. for the two commits under the main commit, change 'pick' to 's' (for 'squash'):

screenshot 2015-01-30 15 01 29

  3. save and quit out of the editor
  4. you'll get bounced to a new editor session. fixup the commit message here:

starts like
screenshot 2015-01-30 15 04 14

make it something like
screenshot 2015-01-30 15 05 05

  5. save and quit out of that editor. feel your git neckbeard grow thick and mighty. rejoice.
screenshot 2015-01-30 15 05 45

6a68:
oh, forgot the last step: force push (git push -f) to this branch, then hit that big green button

6a68:
@mostlygeek we should talk about how you guys could use ruby yaml tools to handle non-flat configs. It seems kinda confusing to me that we're using puppet (a ruby tool), and we're feeding it yaml configs (the ruby conf file format), but we're not using a ruby one-liner to transform the configs. cc @nchapman

pdehaan:
On the topic of flat configs, doesn't [**fxa-content-server**'s config](https://github.com/mozilla/fxa-content-server/blob/master/server/lib/configuration.js) use the same/similar nesting?

pdehaan:
Anybody else care to re-re-review before I press the green button?
I ripped out the SES transporter and replaced with an SMTP transporter, at the request of @mostlygeek.

Otherwise, we can merge now and refactor the code next week.

6a68:
sure, I'll do a final skim

6a68:
lgtm, fire when ready

6a68:
ok, I'll hit that button ^_^

PR #233, opened 2015-01-30
feat(api): add url2png to provide screenshots

6a68:
@nchapman r?

- add url2png package, update npm shrinkwrap
- add url2png keys to config
- add screenshotUrl to visit, visits, and search items
  - implemented as a view-level transform (not in the DB)
  - this patch introduces some duplication, but I want to separate the "get it working" task from the "make it pretty" refactoring task
- also fix a stray SQL error in `server/models/visits` that I missed with yesterday's giant PR

### DIFF
--- a/config/local.json.example
+++ b/config/local.json.example
@@ -23,6 +23,10 @@
     "apiKey": "",
     "enabled": true
   },
+  "url2png": {
+    "apiKey": "",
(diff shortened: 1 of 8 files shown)


nchapman:
Works great! Woot! :+1:

PR #234, opened 2015-01-30
feat(front-end): Real styles and real data displayed for visits

nchapman:
Here's the first pass at getting real data from the API displayed and styled on the front-end. There's still a lot more polish to do, color variables to extract, etc, but I wanted to get this landed ASAP. Look forward for a follow on PR that will hook up search, user's avatar, site screenshots, delete button, and soo *much* **more**!

@6a68 r?


app/scripts/views/visits/item.js
```diff
@@ -16,9 +19,22 @@ define([
...
 
+    getContext: function () {
+      return this.presenter;
+    },
+
+    afterRender: function () {
+      // add size class
+      this.$el.addClass(this.presenter.getSizeClassName());
+      // add image position class
+      this.$el.find('.image').addClass(_.sample(['left', 'right']));
```
nchapman: This is gross. It'll be back in the template on the next pass.

grunttasks/usemin.js
```diff
@@ -19,10 +19,18 @@ module.exports = function (grunt) {
     options: {
       assetsDirs: [
         '<%= staticPath %>'
-      ]
+      ],
+      patterns: {
+        js: [
+          [/(\/images\/.*?\.png)/gm, 'Update the JS to reference revved images']
```
nchapman: This could be more robust in the future but covers our needs for now.

app/scripts/presenters/visit_presenter.js
```diff
@@ -0,0 +1,102 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'underscore'
+], function (_) {
+  'use strict';
+
+  var MINIMUM_IMAGE_ENTROPY = 1.75;
+  var MINIMUM_IMAGE_WIDTH = 270;
+  var EXTRACTED_IMAGE_DENY_PATTERN = /github.com|(\..{2,3}\/$)|([&?]q=)|([&?]page=\d+)/;
```
pdehaan: Not sure if you want to leave a quick comment for future travelers (or mere QE) who may be scratching their heads over all the OR statements in the reg' expression (and why github seems to be special).

app/scripts/views/visits/item.js
nchapman: Mildly related... I want the same items to always be on the left and right so that they don't reshift when you press the back button. I want to derive something from the visit/user_page so that it's consistent. I was thinking of doing `(title.length % 2) ? 'left' : 'right'`. Too dumb?

app/scripts/presenters/visit_presenter.js
nchapman: Yeah agreed it's bad. I'm going to take another pass at that in the next PR so I'll fix it up then.

app/scripts/presenters/visit_presenter.js
```diff
@@ -0,0 +1,102 @@
...
+
+  function VisitPresenter (visit) {
```
nchapman: This whole file may go away at some point. Most of these decisions/features will end up implemented on the server side.

app/scripts/presenters/visit_presenter.js
```diff
@@ -0,0 +1,102 @@
...
+      return this.extractedImageWidth && this.extractedImageWidth >= MINIMUM_IMAGE_WIDTH;
+    },
+
+    getInterestingness: function () {
+      if (!this.interestingness) {
+        var value = 1.0;
+
+        // Positives
+        if (this.hasValidExtractedImageUrl()) {
+          value *= 1.3;
```
nchapman: All these multipliers are arbitrary. We'll come up with something more meaningful at some point.

grunttasks/usemin.js
```diff
@@ -6,7 +6,7 @@ module.exports = function (grunt) {
   'use strict';
 
   grunt.config('useminPrepare', {
-    app: {
+    html: {
```
nchapman: I didn't mean to leave this change :/. Maybe it's better?

app/scripts/views/visits/item.js
6a68: Seems like a reasonable little hack :-) We can always change it later

app/scripts/presenters/visit_presenter.js
```diff
@@ -0,0 +1,102 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
pdehaan: On the topic of filenames... I noticed some recent refactorings from underscores to dashes, not sure if we have a coding style guide yet or if it's worth worrying about early, or if we want to change these to "visit-presenter.js" to match "/server/work-queue/jobs/extract-page.js". /cc @6a68

Possibly same question applies to filenames like "app/images/icon-favicon_default@2x.png" above (dashes and underscores). I can try doing some shell-fu to see if I can find the incantation for listing all filenames minus special folders like .git, node_modules, app/bower_components, etc...

6a68:
I'm all for just merging it and cleaning up async. @nchapman do you want specific feedback on anything in particular here?

app/scripts/presenters/visit_presenter.js
nchapman: I like underscores in filenames.

app/scripts/templates/visits/item.html
```diff
@@ -1,4 +1,12 @@
-<div class="visit">
+<div class="image" style="background-image: url({{extractedImageUrl}});"></div>
```
pdehaan: Instead of ``? I think the CSS approach will prevent me from right-click+saving images...
Or not sure if there is a method we could use for lazy-loading images, if we didn't want to fire off 25+ network requests immediately (or whatever).

app/scripts/presenters/visit_presenter.js
6a68: I eventually settled on using dashes on the server, but these kind of niceties can wait till the initial dust settles ^_^

app/scripts/templates/visits/item.html
nchapman: Yeah we're doing this because we're using CSS to "crop" the images using background cover.

app/styles/_base.scss
```diff
@@ -1,14 +1,8 @@
 @import '../bower_components/normalize-scss/normalize';
-@import url(http://fonts.googleapis.com/css?family=Fira+Sans:400,700,400italic,700italic);
+@import url(https://fonts.googleapis.com/css?family=Fira+Sans:400,500,700,400italic,700italic);
```
pdehaan: change to protocol agnostic "//fonts.googleapis.com/"... ?

nchapman: Yeah, I didn't at first because I was afraid sass would think it was a comment. Seems to be fine. Pushing now...

app/scripts/templates/visits/item.html
```diff
@@ -1,4 +1,12 @@
...
+<div class="text">
+  <div class="provider">
+    <img class="favicon" src="{{faviconUrl}}" />
```
pdehaan: Not sure if you want to add `alt="{{extractedProviderName}}"` for screen readers or whatevers.

nchapman:
I'll merge this once Travis gives his blessing.

app/styles/modules/_global_header.scss
```diff
@@ -0,0 +1,39 @@
+header {
+  background: #4d4d4d;
```
pdehaan: @johngruen is the official Sass owner, but I'll let you guys decide if you want to go with inline values versus declaring "allowed" colors in the <https://github.com/mozilla/chronicle/blob/master/app/styles/_variables.scss> file.

app/styles/modules/_global_header.scss
```diff
@@ -0,0 +1,39 @@
...
+  top: 0;
+  z-index: 100;
+
+  .row-inner {
+    height: 65px;
+  }
+
+  h1 {
+    @include vertical-align;
+    color: #fff;
```
pdehaan: The _variables.scss mixin has a `$white` variable for this. :shrug:

app/styles/modules/_global_header.scss
nchapman: Yeah agreed, I plan on extracting colors. We'll have to decide if we want to extract every color or just colors we used repeatedly.

app/styles/_mixins.scss
```diff
@@ -0,0 +1,34 @@
+@mixin vertical-align {
+  position: relative;
+  top: 50%;
+  transform: translateY(-50%);
+}
+
+@mixin two-by-three($width) {
+  height: $width * 0.667;
```
pdehaan: I don't imagine this would stretch anything, but should this be `max-height` here?

app/scripts/presenters/visit_presenter.js
pdehaan: I never said it was bad, i just said I couldn't read it and understand what it was trying to do.

PR #235, opened 2015-01-31
fix(db): improve elasticsearch query

### LINKED ISSUE #227: improve search results
Let's tweak the elasticsearch query to search on multiple fields.
It might also be useful to keep the keywords and use those to boost matches - we could store them as a json array in postgres, i.e. an opaque json blob, and actually use them in elasticsearch.

6a68:
Fixes #227 

Stole @nchapman's v0 search query, let's rock!

### DIFF
--- a/server/models/visits.js
+++ b/server/models/visits.js
@@ -75,29 +75,25 @@ var visits = {
       size: count,
       body: {
         query: {
-          match: {
-            extractedContent: searchTerm
+          filtered: {
+            query: {
+              multiMatch: {
+                query: searchTerm,
+                fuzziness: 'AUTO',
+                operator: 'and',
+                fields: [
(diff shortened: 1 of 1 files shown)


server/models/visits.js
```diff
@@ -75,29 +75,25 @@ var visits = {
...
+          filtered: {
+            query: {
+              multiMatch: {
+                query: searchTerm,
+                fuzziness: 'AUTO',
+                operator: 'and',
+                fields: [
+                  'extractedTitle',
+                  'extractedContent',
+                  'providerDisplay',
```
nchapman: Do these need the `extracted` prefix?

6a68: doh, yup

nchapman:
:+1:

PR #236, opened 2015-01-31
chore(sass): tweaking some styles per scss-lint feedback

pdehaan:
For more info on **scss-lint**, see https://github.com/causes/scss-lint.
It requires Ruby, so it doesn't make sense to add this to any "lint" workflows, but it is available as a npm script convenience method (`npm run scss-lint`) if you already have **scss-lint** installed locally.

Current output is this (but I don't agree with merging):

```
app/styles/modules/_visit.scss:115 [W] Merge rule `&.small + &.small` with rule on line 94
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "lint": "grunt lint",
     "outdated": "npm outdated --depth 0",
     "postinstall": "bower update --config.interactive=false -s",
+    "scss-lint": "scss-lint app/styles || true",
     "shrinkwrap": "npm shrinkwrap --dev && npm run validate-shrinkwrap",
     "start": "grunt serve",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -68,6 +68,7 @@
     "lint": "grunt lint",
     "outdated": "npm outdated --depth 0",
     "postinstall": "bower update --config.interactive=false -s",
+    "scss-lint": "scss-lint app/styles || true",
```
pdehaan: The `|| true` hackery is because I didn't really care about [exit status codes](https://github.com/causes/scss-lint#exit-status-codes).

nchapman:
Cool thanks @pdehaan!

PR #239, opened 2015-01-31
fix(server): Fixes FxA sign in

nchapman:
Seems like the fix for FxA sign in was just renaming the fetched user object. I'm going to merge this right away so I can fix the dev server unless someone is really quick to object :smile:.


server/bell-oauth-profile.js
```diff
@@ -46,7 +46,8 @@ function profile (credentials, params, get, profileCb) {
           }
           log.verbose('created new user ' + data.email);
           credentials.profile = {userId: data.uid, isNewUser: true};
-          queue.sendWelcomeEmail({email: data.email});
+          // FIX: This doesn't appear to be here. Disabling for now.
```
nchapman: @pdehaan check this out when you have a minute.

PR #242, opened 2015-02-01
feat(front-end): Adds image proxy via embedly

nchapman:
This adds Embedly Display as an image proxy on the front-end. This should fix any mixed content warnings over https and does a little work to serve retina images when appropriate (could be brittle). This also replaces our use of CSS to "crop" images using background cover. This makes images actual img tags instead of scumbag background images.

@6a68 @pdehaan r?

### DIFF
--- a/app/scripts/config.js
+++ b/app/scripts/config.js
@@ -7,8 +7,8 @@ define([
   'use strict';
 
   var config = {
-    oauth: {
-      authEndpoint: '{%= config.get("server.oauth.authEndpoint") %}'
+    embedly: {
+      apiKey: '{%= config.get("embedly.apiKey") %}'
     }
   };
 
--- a/app/scripts/lib/image_proxy.js
+++ b/app/scripts/lib/image_proxy.js
@@ -1,0 +1,49 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'underscore',
+  'jquery',
+  'config'
+], function (_, $, config) {
+  'use strict';
+
+  var EMBEDLY_DISPLAY_ENDPOINT = '//i.embed.ly/1/display';
+
+  // TODO: move this somewhere generic when it's needed in more than one place
+  var IS_RETINA = window.devicePixelRatio >= 1.3;
+
+  var imageProxy = {
+    display: function (url) {
+      return this._buildUrl('', { url: url });
+    },
+
+    crop: function (url, width, height) {
+      return this._buildUrl('/crop', { url: url, width: width, height: height });
+    },
+
+    resize: function (url, width, height, grow) {
+      return this._buildUrl('/resize', { url: url, width: width, height: height, grow: grow });
+    },
+
+    _buildUrl: function (action, options) {
+      // add api key, retinify dimension options, and convert to query params
+      var params = $.param(_.extend({ key: config.embedly.apiKey }, this._retinafyDimensions(options)));
+
+      return EMBEDLY_DISPLAY_ENDPOINT + action + '?' + params;
+    },
+
+    // Increases the size of images to account for retina displays
+    _retinafyDimensions: function (options) {
+      if (IS_RETINA && options.width && options.height) {
+        options.width = options.width * 2;
+        options.height = options.height * 2;
+      }
+
+      return options;
+    }
+  };
+
+  return imageProxy;
+});
--- a/app/scripts/presenters/visit_presenter.js
+++ b/app/scripts/presenters/visit_presenter.js
@@ -3,12 +3,20 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 define([
-  'underscore'
-], function (_) {
+  'underscore',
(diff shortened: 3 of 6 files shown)


app/scripts/config.js
```diff
@@ -7,8 +7,8 @@ define([
   'use strict';
 
   var config = {
-    oauth: {
-      authEndpoint: '{%= config.get("server.oauth.authEndpoint") %}'
+    embedly: {
+      apiKey: '{%= config.get("embedly.apiKey") %}'
```
6a68: is this safe to send to the client long-term?

nchapman: It's how they suggest doing it... We can restrict to certain IPs and referrers in account settings. We could also add our own endpoint that proxies back to theirs... so many proxies!

app/scripts/presenters/visit_presenter.js
```diff
@@ -22,7 +30,15 @@ define([
...
-      return this.hasValidExtractedImageUrl() ? this.extractedImageUrl : this.screenshot_url;
+      var url = this.hasValidExtractedImageUrl() ? this.extractedImageUrl : this.screenshot_url;
+
+      // it would be smarter to only crop screenshots if we're not in retina, but it's easier
+      // to let the imageProxy just handle it for now.
+      return imageProxy.crop(url, IMAGE_DISPLAY_WIDTH, IMAGE_DISPLAY_HEIGHT);
+    },
+
+    imagePosition: function () {
+      return (this.title.length % 2) ? 'left' : 'right';
```
nchapman: This could use an explanation...

app/scripts/lib/image_proxy.js
```diff
@@ -0,0 +1,49 @@
...
+  'config'
+], function (_, $, config) {
+  'use strict';
+
+  var EMBEDLY_DISPLAY_ENDPOINT = '//i.embed.ly/1/display';
+
+  // TODO: move this somewhere generic when it's needed in more than one place
+  var IS_RETINA = window.devicePixelRatio >= 1.3;
+
+  var imageProxy = {
```
6a68: It looks like only the `crop` method is called in this patch. Ordinarily, I'd suggest maybe getting rid of the unused other parts of the API, but this file's whole existence seems a bit awkward/suspect - really seems like the server should be doing the image sizing / proxy URL generation for the front-end. So, since this smells like a short-term fix, I think it's fine as-is.

app/scripts/presenters/visit_presenter.js
6a68: seems readable to me *shrug*

6a68:
@nchapman r+, let's keep things moving and discuss followups in sprint planning today ^_^

app/scripts/lib/image_proxy.js
nchapman: Yeah I was playing with all of them. I think I should either get rid of the unused ones or implement all of the methods that the display endpoint supports.

app/scripts/config.js
6a68: Aha, cool. I should look at their docs some more :blush:

app/scripts/presenters/visit_presenter.js
pdehaan: I think the _code_ is readable. I'd question the _logic_ of alternating the floats of the images though. Seems like it'll lead to an inconsistent UI which can be a bit of a drag when things appear randomly shuffled. Hurts my muscle memory.

app/scripts/presenters/visit_presenter.js
```diff
@@ -3,12 +3,20 @@
...
-], function (_) {
+  'underscore',
+  'lib/image_proxy'
+], function (_, imageProxy) {
   'use strict';
 
   var MINIMUM_IMAGE_ENTROPY = 1.75;
-  var MINIMUM_IMAGE_WIDTH = 270;
+  var IMAGE_DISPLAY_WIDTH = 270;
+  var IMAGE_DISPLAY_HEIGHT = 180;
```
pdehaan: Not sure if we use these constants elsewhere, or if they're candidates for moving into the Convict config file. I see that we're also hardcoding a "270px" in the Sass files below, so convict wouldn't help with that.

pdehaan: Without digging too far into the code, not sure if I'm reading the logic:
Should these be equal values?
```js
var IMAGE_DISPLAY_HEIGHT = 180;
```

vs

```css
# app/styles/modules/_visit.scss

// Size variations
&.medium {
  .image img {
    @include two-by-three(160px);
  }
}
```

Should `IMAGE_DISPLAY_HEIGHT` be 160px, or should it be `@include two-by-three(180px);` or are these two completely different things?

nchapman: 160px is the width, and it will calculate the height in that function. The largest we ever use is 270px wide and in this case we let the browser scale the images down to 160px.

app/scripts/presenters/visit_presenter.js
nchapman: Yeah I think it's readable too, but the logic behind why we're using title this way may not be clear (it's so an item never changes sides). The reason for alternating them is a design decision to lend visual interest and rhythm. We're already changing image sizes based on "interestingness", so the button targets will be moving anyway.

PR #266, opened 2015-02-04
feat(front-end): Adds user info data with avatar to header

nchapman:
This should be merged after #265.

* * *

Adds user info box to the header. Looks like this right now:

screen shot 2015-02-03 at 8 15 24 pm

Obviously not awesome yet, but ready for @johngruen to style up. We can have that happen in this branch or just merge this and clean up later. Either works for me.

@6a68 @pdehaan r?

PR #265, opened 2015-02-04
feat(front-end): Working search box and results

nchapman:
Working search results are here! Lots of new views. The header is now a little stack of specialized views. The visit presenter has been reworked as a user page presenter. Tried to share a reasonable amount of code between visits and search. I avoided making many scss changes since @johngruen is working on that in a separate branch.

@6a68 @pdehaan r?

PR #267, opened 2015-02-04
feat(front-end): Endless scrolling of visits

nchapman:
This should be merged after #265 and #266.

* * *

Here's a take on endless scrolling for visits. I wish it could be simplified further, but it at least does the job for now. It doesn't do anything to keep the back button working so we'll need to fix that in another pass (or target=blank all the things).

@6a68 let me know what you think about this... We can do something simpler if you think that's better.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -21,6 +21,7 @@
   "globals": {
     "define": false,
     "Promise": true,
-    "window": true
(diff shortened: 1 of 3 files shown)


6a68:
@nchapman dude, let's just land it and see how we like it!

PR #277, opened 2015-02-04
style(footer): dejank the footer

johngruen:


app/index.html
```diff
@@ -10,9 +10,18 @@
...
+        <h1>Chronicle</h1>
+        <h2>Welcome home</h2>
+        <a href="/auth/login">Sign in</a>
+      </div>
+    </div>
+    <footer id="global-footer" class="row">
+      <div class="row-inner">
+        <p>&copy; 2015 &bull; <a href="http://mozilla.org/" target="_blank">Mozilla</a></p>
+      </div>
+    </footer>
```
pdehaan: Since this footer seems to be shared between the app/index.html and app/app.html, not sure if it can be recycled similar to the `<!-- view: global_header/global_header -->` for the app.html global header.

pdehaan:
Apart from the footer duplication (which is a solid "whatever", at best), this looks good to me!
I thought that we used to be able to hover over the gravatar to sign out, but I can't seem to get that working locally. I don't think it's related to this PR since I can't get it working in master either. :/

PR #281, opened 2015-02-04
Digging into db problems

6a68:
Adding a bit of logging to try to figure out what's up with dev.


6a68:
I"ll just debug directly on dev and submit this if it fixes anyting ^_^

Issue #282, opened 2015-02-05
ensure promise failure results in DB connection cleanup


6a68:
it looks like visit model failure is not cleaning up DB connections. Not sure why.

6a68:
I think this was fixed by #283, and given time constraints, we're going to have to not be proactive about this kind of code quality audit for now. Closing.

PR #283, opened 2015-02-05
Properly delete visits

6a68:
@nchapman @pdehaan r?

TL;DR: visit deletion was busted because of the bug fixed by this patch.

The visit.delete model code was expecting an integer count to be returned by postgres, but `{count: <int>}` was getting returned instead. By the laws of javascript insanity, the code then declared shenanigans, told itself that `count > 0` was definitely not true, and, thus, decided that it was time to delete the user_page, when in fact visits still existed that corresponded to said user_page. That led to a postgres error being thrown and unhandled (because the visits table references the user_pages table as a foreign key, and we do _not_ cascade deletes, in order to avoid accidentally nuking data thanks to exactly this kind of bug!). I suspect/hope the DB connection leakage was happening when that uncaught error nuked the process, since systemctl merrily spun up a new one in response.

I've filed #282 to figure out wtf is going on with our error handling cleanup in the models; the problem may simply be that no graceful shutdown happened, so the db connection was leaked. Still seems highly suboptimal handle uncaught exceptions in this way, since, uh, I expect this won't be the last one that leaks :-\

To be continued, I guess.

I've also fixed a typo that was causing jshint to barf.

The bug fix commit fixes the problems we're seeing in the logs on dev, so (fingers crossed) dev should be solid again after landing this.

### DIFF
--- a/server/db/postgres.js
+++ b/server/db/postgres.js
@@ -42,7 +42,7 @@ var postgres = {
   //
   // query := a prepared query string
   // params := an array of the query params in correct order
-  query: function query(query, params) {
+  query: function query(queryString, params) {
     var _defer = Q.defer();
     var formatted;
     // force promises to eventually resolve
@@ -53,7 +53,7 @@ var postgres = {
         log.trace(err);
         return _defer.reject(err);
       }
-      client.query(query, params, function onQueryResponse(err, results) {
+      client.query(queryString, params, function onQueryResponse(err, results) {
         if (err) {
           log.warn('postgres query failed');
           log.trace(err);
--- a/server/models/visit.js
+++ b/server/models/visit.js
@@ -132,22 +132,28 @@ var visit = {
         [userId, userPageId]);
       })
       .fail(visit._onRejected.bind(visit, name + ' failed to count visits having a page', cb))
-      .then(function(count) {
-        _verbose('count of remaining visits with page gives us: ' + JSON.stringify(count));
-        // assuming count is int
-        if (count > 0) {
+      .done(function(result) {
+        var count = result && parseInt(result.count, 10);
+        // non-numbers will be coerced to NaN by parseInt, which is typeof 'number', lol
+        // note that if result is undefined, count will be, too, and isNaN(undefined) is true.
+        if (isNaN(count)) {
+          log.warn('attempting to count remaining visits sharing a user_page failed to return a numeric count, count was ' + count);
+          visit._onRejected(name + ' failed to return a count of visits sharing a user_page', cb, null);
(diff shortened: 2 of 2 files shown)


6a68:
I should also mention that this branch is currently on dev. If it's still wonky or busted, maybe I need to dig some more and so we shouldn't land this branch.

pdehaan:
I'm still seeing some issues with deleting items on the dev server. Basically my FF crashed this morning and came back with about 40 results of "New Tab" which got pushed into my Chronicle feed. So I'm trying to go through and delete all those entries manually in Chronicle and my Developer Tools ends up spitting up this message:

> _"Uncaught Error: A `url` property or function must be specified"_

The stack trace is pretty useless, but seems to point to a reference to `t.extend.destroyModel` in [/app/scripts/views/visits/item.js:40](https://github.com/mozilla/chronicle/blob/98d26557937193e8515ccfa267bdb6a905a64ac8/app/scripts/views/visits/item.js#L40)

```js
36:    destroyModel: function (event) {
37:      event.preventDefault();
38:
39:      if (window.confirm('Destroy this visit?')) {
40:        this.model.destroy();
41:
42:        this.$el.fadeOut(function () {
43:          this.remove();
44:        }.bind(this));
45:      }
46:    }
```
Not sure if we need a `try {} catch (err) {}` around that to see why it's sad. Or why it seems to work some times and then begins failing.

chronicle_and_fdc55c2c022172eb5d9c37b9f404fc70_json_ _embedly-cli_and_config_get_txt_ _chronicle

pdehaan:
Long story short, it's probably me who keeps murdering the poor dev server with delete-fest 2014.

6a68:
@pdehaan That actually makes sense - I've seen a lot of similar bugs lately where Nightly loses the URL when opening a new tab, or detaching a tab from a tab group, or after restarting - sometimes the page loads but the URL is blank, sometimes the page just doesn't load at all. If the addon says, "here you go, server, here are 20 visits to 'about:blank' at the exact same time", then yeah, things are gonna get weird.

6a68:
Well, let's merge this puppy and see what happens :-)

pdehaan:
Tested in Firefox a few mins ago and I'm seeing this now (when deleting my 20-30 "New Tab" items):
chronicle_and_javascript_-_put_item_on_dynamodb_table_using_aws_sdk_for_node_js_-_stack_overflow

The good news is that they are removed from the DOM after I press delete, and seemingly are still gone after I refresh the page, so it's just a noisy console.

Not specific to "New Tab", as I can reproduce when deleting boring Google search history items or the "Chronicle" page items.

PR #286, opened 2015-02-05
style(user-menu): make user menu stylish

johngruen:


app/styles/_state.scss
```diff
@@ -10,11 +10,17 @@
   color: $warning;
 }
 
+.is-triggered {
+  background: linear-gradient(to bottom, #060606 0%, #2d2d2d 100%);
```
pdehaan: Not sure about opinions on inlining colors here vs using variables. Your call.

johngruen: @pdehaan: let's leave these be for now, i think color management is going to require a bit more thought.

pdehaan:
This looks great locally, and scss-lint loves it! :heart_eyes: 

Only idea I had was if we'd need some `:hover` state to menu item links to give a bit of user feedback that they are hovering over a link.

I used something like this locally, but it isn't exactly pretty:
```css
    a {
      color: $dark-grey;
      display: block;
      padding: $default-gap 0 0 0;
      text-decoration: none;

      &:hover {
        color: $black;
        text-decoration: underline;
      }
    }
```

chronicle

pdehaan:
Tested in Chrome and Firefox and looks great. :+1: from me!

Here's what it looks like in Firefox on a ridiculously small window width (or possibly a mobile browser). Not sure if we want to remove that black background at smaller breakpoints as it gets a bit skewed:

chronicle_and_breaking_news__u_s___world__weather__entertainment___video_news_-_cnn_com

johngruen:
@pdehaan: yeah might as well add that for now. @bryanbell can put the final behavior in his spec

pdehaan:
I think this looks great! :clap: 
Feel free to merge when you're ready.

nchapman:
I'm taking a look right now too. Give me a minute to catch up :)

johngruen:
@nchapman: don't merge yet, fixing peter's quibbles

pdehaan:
On the topic of css color refactorings... One tool I've used in the past is [**css-colorguard**](https://github.com/SlexAxton/css-colorguard).

Here's the output against this PR and my non-minified "compiled.css" file (using default settings):

```
$ colorguard --file public/styles/compiled.css

Collision: #060606, #000000
  - #060606 (#060606) [line: 15] is too close (0.9483523836028054) to #000 (#000000) [line: 131, 484, 523, 549]
Collision: #2d2d2d, #333333
  - #2d2d2d (#2d2d2d) [line: 15] is too close (1.9180385351876015) to #333333 (#333333) [line: 360, 440, 518]
Collision: #f2f2f2, #ffffff
  - #f2f2f2 (#f2f2f2) [line: 359] is too close (2.6306067895807894) to #ffffff (#ffffff) [line: 397, 411, 424, 455, 462, 480, 500, 547]
Collision: #f2f2f2, #f5f5f5
  - #f2f2f2 (#f2f2f2) [line: 359] is too close (0.6188106551513192) to whitesmoke (#f5f5f5) [line: 599]
Collision: #ffffff, #f5f5f5
  - #ffffff (#ffffff) [line: 397, 411, 424, 455, 462, 480, 500, 547] is too close (2.011983879269603) to whitesmoke (#f5f5f5) [line: 599]
```

nchapman:
The menu doesn't currently close when you click on the body. Here's some code I used in another project that might get you started...

```js
$('body').click(this.bodyClicked.bind(this));

...

function bodyClicked(event) {
  var $target = $(event.target);
  var $statusMenu = $('.status-menu');

   if ($statusMenu.is(':visible') && !$target.is('.status-menu, .status-menu-button') && !$target.closest('.status-menu').length) {
    $statusMenu.fadeOut(175);
  }
}
```

Not great code but might help. You'll need to unbind the click beforeDestroy in the view.

nchapman:
Firefox has a border:

screen shot 2015-02-05 at 10 26 46 am

Chrome doesn't have a border:

screen shot 2015-02-05 at 10 26 56 am

pdehaan:
Interesting... it looks like clicking the "Sign Out of Chronicle" doesn't work.
Clicking the link changes the URL to http://localhost:8080/#auth/logout but nothing seems to happen and the menu seems to stay open.

Not sure _why_ that's happening though, since your link looks kosher to me:
```html
<li><a href="/auth/logout">Sign Out of Chronicle</a></li>
```

But I can file a new issue after we merge this and we can make @nchapman roll up his sleeves and figure it out. I vaguely recall some magic code in his Backbone walkthrough yesterday which was using some regex or something to look at the URL fragments. Maybe that is injecting some "#" in the URL.

nchapman:
Border around the avatar seems a little crunch. Mock seems to have a shadow too.

screen shot 2015-02-05 at 10 29 11 am

vs

screen shot 2015-02-05 at 10 31 57 am

nchapman:
A few more notes...

- Seems like the border radius on the menu is smaller in the mock
- It's missing the divider line between the two sections
- We obviously need some icon assets, but those can come later.

This is looking really good @johngruen!

johngruen:
@nchapman: okay will fix this crapola

app/styles/modules/_user_info.scss
```diff
@@ -0,0 +1,86 @@
...
+      display: block;
+      padding: $default-gap 0 0 0;
+      text-decoration: none;
+      
+      &:hover {
+        text-decoration: underline;
+      }
+    }
+
+    + .chunk {
```
pdehaan: Me thinks `.chunk` should be `.menu-chunk` here.
Changing that gives me:
chronicle

johngruen: @pdehaan, yep, forgot to change that one

pdehaan: I tried to use the glorious shorthand of `+ & {` but it seems that the `&` translates to the entire selector name so it was generating `#user-info .menu-chunk + #user-info .menu-chunk` which was blowing up.

From the docs:
> _`&` will be replaced with the parent selector as it appears in the CSS. This means that if you have a deeply nested rule, the parent selector will be fully resolved before the `&` is replaced._
> &mdash; via http://sass-lang.com/documentation/file.SASS_REFERENCE.html#parent-selector

pdehaan:
Sorry, one last thing (and let me know if you want a separate bug for this), but we may need to play around with this a bit if the user has an offensively long email address. I was playing with it locally in the developer tools, but it seems like the address bleeds outside of the menu. I can toy around a bit more with the convict config (or wherever that email address is coming from) and see if the behavior is any different:

chronicle

johngruen:
@pdehaan added a general utility class to solve the overflow thing

PR #289, opened 2015-02-06
fix(front-end): Fixes endless scroll in Firefox

### LINKED ISSUE #285: Visits don't endlessly scroll on Firefox
Only the first page is displayed. It's probably related to how we're calculating the distance from the bottom.

nchapman:
Fixes #285.

@pdehaan r?

### DIFF
--- a/app/scripts/views/visits/index.js
+++ b/app/scripts/views/visits/index.js
@@ -97,7 +97,7 @@ define([
     },
 
     _getPixelsFromBottom: function () {
-      return $(document).height() - $('body').scrollTop() - $(window).height();
+      return $(document).height() - $(document).scrollTop() - $(window).height();
     },
   });
 


app/scripts/views/visits/index.js
```diff
@@ -97,7 +97,7 @@ define([
     },
 
     _getPixelsFromBottom: function () {
-      return $(document).height() - $('body').scrollTop() - $(window).height();
+      return $(document).height() - $(document).scrollTop() - $(window).height();
```
nchapman: Should have been this way all along. I must have been tired.

pdehaan:
Tested locally w/ Chrome and Firefox and worked like magic. :+1:

(**PROTIP:** You can set your item your paging count in [/app/scripts/views/visits/index.js:26](https://github.com/mozilla/chronicle/blob/d05e4b6cfda36e4a3fb79f64ee275ddc24b7d4e8/app/scripts/views/visits/index.js#L26).)

PR #290, opened 2015-02-06
feat(db): add migrations support

### LINKED ISSUE #92: DB migrations
We will need these when we go to production. Figure them out.

6a68:
Fixes #92

@nchapman @pdehaan r?

Seems to be working locally for me. If you want to play around with migrations to different levels, we currently have 0, 1, and 2; just do `./bin/migrate.js 0` or whichever int you'd like.

This does not solve the reindexing problem for elasticsearch, that will be in followup issue #291.

### DIFF
--- a/bin/migrate.js
+++ b/bin/migrate.js
@@ -1,0 +1,35 @@
+#!/usr/bin/env node
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// use this script to call our pg-patcher wrapper via the command line.
+// TODO add 'next' and 'prev' convenience methods?
+// TODO bring back commander, I guess
+
+var fs = require('fs');
+var log = require('../server/logger')('bin.migrate');
+var migrator = require('../server/db/migrator.js');
+var targetLevel = parseInt(process.argv[2], 10);
+
+// if no args were passed, or if the arg wasn't an integer,
+// then, by default, use the highest defined patch level
+if (isNaN(targetLevel)) {
+  files = fs.readdirSync(__dirname + '/../server/db/migrations');
+  var max = 0;
+  files.forEach(function(file) {
+    // assume files are of the conventional form 'patch-n-m.sql'
+    // we want the largest m in the dir
+    // the regex splits the filename into ['patch', 'n', 'm', 'sql']
+    var curr = file.split(/[-\.]/)[2];
+    if (curr > max) { max = curr; }
+  });
+  targetLevel = max;
+}
+migrator(targetLevel)
+  .done(function onSuccess(resp) {
+    log.info('migration succeeded');
+  }, function onFailure(err) {
+    log.error('migration failed: ' + err);
+  });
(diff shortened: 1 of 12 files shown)


6a68:
whoops, sorry travis

bin/migrate.js
```diff
@@ -0,0 +1,35 @@
...
+
+var fs = require('fs');
+var log = require('../server/logger')('bin.migrate');
+var migrator = require('../server/db/migrator.js');
+var targetLevel = parseInt(process.argv[2], 10);
+
+// if no args were passed, or if the arg wasn't an integer,
+// then, by default, use the highest defined patch level
+if (isNaN(targetLevel)) {
+  files = fs.readdirSync(__dirname + '/../server/db/migrations');
```
pdehaan: Not sure if we want to use `path.join()` here (or if we're concerned about potential Win+Mac path issues yet).

server/db/migrations/patch-2-1.sql
```diff
@@ -0,0 +1,6 @@
+-- drop the visits table
+DROP TABLE visits;
+-- drop the user_pages table
+DROP TABLE user_pages;
+-- drop the users table
+DROP TABLE users;
```
pdehaan: Do we need to delete the indexes, or are those deleted with the tables?

server/db/migrations/patch-1-2.sql
```diff
@@ -0,0 +1,66 @@
...
+  extracted_media_duration INTEGER,
+  extracted_media_height INTEGER,
+  extracted_media_html TEXT,
+  extracted_media_type TEXT,
+  extracted_media_width INTEGER,
+  extracted_provider_display TEXT,
+  extracted_provider_name TEXT,
+  extracted_provider_url VARCHAR(2048),
+  extracted_published TIMESTAMPTZ(3),
+  extracted_safe BOOLEAN,
```
pdehaan: Unrelated to life, but me thinks we may need another couple columns for when `extracted_safe` is `false`:

- `extracted_safe_type` &mdash; either "phishing" or "malware".
- `extracted_safe_message` &mdash; a user-readable message in HTML format explaining that the given url is likely malicious.

(per http://embed.ly/docs/api/extract/features/safe)

bin/migrate.js
6a68: yeah, about windows...

6a68: good point though, I'll just use path.join ;-)

server/db/migrations/patch-2-1.sql
6a68: you guessed it. from [the docs](http://media.giphy.com/media/ryOujCIH2dyla/giphy.gif):

> DROP TABLE always removes any indexes, rules, triggers, and constraints that exist for the target table. However, to drop a table that is referenced by a view or a foreign-key constraint of another table, CASCADE must be specified.

Rather than rely on CASCADE, I just delete tables in reverse order of dependency, in hopes of making the migrations more easily scannable.

server/db/migrations/patch-1-2.sql
6a68: Rather than show the user embedly's safe_message, we will probably rely on some UI that, say, either hides or fades back or puts a warning sign next to such entries, assuming they ever exist.

I'll file a followup bug to handle this edge case a little bit later.

bin/migrate.js
6a68: actually, I'll file a followup bug to audit our codebase for windows compatibility issues

6a68: filed #294

6a68:
ok, thanks very much for the review, @pdehaan! merging.

Issue #280, opened 2015-02-04
Delete visit has a REALLLLLLLY generous hit target.


pdehaan:
Probably a temporary issue, but I'm a firm believer in "see something, file something".

It seems that the entire `<li>` works as a click target, and extends horizontally across the entire footer area.

![chronicle](https://cloud.githubusercontent.com/assets/557895/6050213/a91e02ea-ac72-11e4-806b-92f1a6789886.png)

It also seems I can fix it by changing the destroy HTML to `<a href="#" class="destroy">Destroy</a>` so now only the text triggers the super awesome `confirm()` item delete UX. I think I traced it back to a recent change, so not sure why we made this fix:
https://github.com/mozilla/chronicle/pull/272/files#diff-589ed4b6805c6296d743549e3301977eR11

nchapman:
This shouldn't issue for long. I moved the class up to the `li` because we hide that on search results. Once this is styled, I think the click target will be appropriately sized.

nchapman:
More thoughts: we can move the click event down to the anchor, but I figured we might end up changing which elements we use here anyway.

johngruen:
Fixed by #292

nchapman:
This appears to still be an issue in Firefox and Chrome.

nchapman:
Fixed in #306.

PR #292, opened 2015-02-06
styles(destroy): iconify destroy button

johngruen:
@nchapman @pdehaan: this directly addresses #280, but it's also setting up formally styling the .actions area. This PR isn't intended to totally match @bryanbell's mockups, just an incremental improvement.


app/styles/modules/_actions.scss
```diff
@@ -0,0 +1,32 @@
...
+      @include hidpi-background-image(icon-delete-active, 20px 20px);
+    }
+  }
+}
+
+.medium .actions,
+.large .actions {
+  clear: both;
+  display: block;
+}
```
pdehaan: FWIW, I think you can tweak this to (or not, your call):

```css
.actions {
  ...

  .medium &,
  .large & {
    clear: both;
    display: block;
  }
}
```

app/styles/modules/_user_page.scss
```diff
@@ -3,8 +3,8 @@
   border: 1px solid $baby-grey;
   // negative spread keeps the shadow from leaking out the sides
   box-shadow: 0 2px 4px -3px rgba(0, 0, 0, 0.14);
-  overflow: auto;
   padding: $double-gap;
+  overflow: auto;
```
pdehaan: A-Z here and in `.small` and in the `&:after` on L122.

pdehaan:
Looks good locally!
Curious, why not use an SVG font like http://fortawesome.github.io/Font-Awesome/icon/trash-o/ or https://icomoon.io/#linearicons

nchapman:
> why not use an SVG font

SVG/font icons are usually not sharp at small sizes in standard def. It's also a lot more trouble to have to translate their visual style into CSS rather than just exporting new versions as the style evolves.

pdehaan:
Oh weird, I thought we switched to Fontello for FxA content server:
[**mozilla/fxa-content-server** /app/styles/fontello.css](https://github.com/mozilla/fxa-content-server/blob/master/app/styles/fontello.css) (see https://github.com/mozilla/fxa-content-server/pull/1405)

TL;DR

nchapman:
I'll just leave this here: https://github.com/mozilla/fxa-content-server/blob/master/app/styles/fontello.css#L11. If one cares about pixel perfection (we do), fonts are hard.

johngruen:
@pdehaan @nchapman: we should at least spritify these things eventually, once we get the final assets

nchapman:
> we should at least spritify these things

@johngruen I talked to @bryanbell about that last night. He's on it.

nchapman:
:+1:

pdehaan:
Saw this on the weekend: https://www.npmjs.com/package/grunt-spritesmith
Not sure if having Grunt maintain our spritesheet would be overkill.

PR #293, opened 2015-02-06
style(breakpoints): add basic breakpoint logic

johngruen:
@nchapman: This seems like a good solution...what do you think?

### DIFF
--- a/app/app.html
+++ b/app/app.html
@@ -15,14 +15,14 @@
       <!-- view: global_header/global_header -->
     </header>
     <div id="container" class="row">
-      <div class="row-inner">
+      <div class="row-inner-break">
         <section id="stage">
           <!-- view: dynamically by router.js -->
         </section>
       </div>
     </div>
     <footer id="global-footer" class="row">
-      <div class="row-inner">
+      <div class="row-inner-padded">
         <p>&copy; 2015 &bull; <a href="http://mozilla.org/" target="_blank">Mozilla</a></p>
       </div>
     </footer>
--- a/app/scripts/templates/global_header/global_header.html
+++ b/app/scripts/templates/global_header/global_header.html
@@ -1,4 +1,4 @@
-<div class="row-inner l-flex-horizontal">
+<div class="row-inner-padded l-flex-horizontal">
   <div id="user-info">
     <!-- view: user_info -->
   </div>
--- a/app/styles/_breakpoints.scss
+++ b/app/styles/_breakpoints.scss
@@ -1,0 +1,19 @@
+// Breakpoint management
+// http://www.sitepoint.com/managing-responsive-breakpoints-sass/
+
+$breakpoints: (
+  small: ('max-width: 520px'),
+  big: ('min-width: 521px')
+);
+
+@mixin respond-to($breakpoint) {
+  @if map-has-key($breakpoints, $breakpoint) {
+    @media (#{map-get($breakpoints, $breakpoint)}) {
+      @content;
+    }
+  }
+  @else {
+    @warn "Unfortunately, no value could be retrieved from `#{$breakpoint}`. "
+        + "Please make sure it is defined in `$breakpoints` map.";
+  }
+}
--- a/app/styles/_variables.scss
+++ b/app/styles/_variables.scss
@@ -39,4 +39,3 @@ $round-radius: 50%;
 //TIMING VARIABLES
 $default: 150ms;
 $fast: $default * 0.5;
-
--- a/app/styles/main.scss
+++ b/app/styles/main.scss
@@ -4,6 +4,7 @@
 
 @import 'variables';
 @import 'mixins';
+@import 'breakpoints';
 @import 'state';
 @import 'base';
 @import 'layout';
--- a/app/styles/modules/_global_header.scss
+++ b/app/styles/modules/_global_header.scss
@@ -13,6 +13,9 @@
     font-size: $large-font-size;
     line-height: $large-font-size;
     margin: 0 $small-gap;
+    @include respond-to(small) {
+      display: none;
+    }
 
     a {
       color: $white;
@@ -22,6 +25,9 @@
 
   #search-box {
     flex-basis: $default-grid * 10;
+    @include respond-to(small) {
+      flex: 1;
+    }
 
     form {
       flex: 1;
@@ -37,23 +43,6 @@
       flex: 1;
       font-size: $default-font-size;
       padding: $small-gap $quad-gap $small-gap $small-gap;
-    }
-  }
-}
-
-@media screen and (max-width: 520px){
-
-  #global-header {
-    h1 {
-      display: none;
-    }
-
-    #user-info {
-      margin-right: $default-gap;
-    }
-
-    #search-box {
-      flex: 1;
     }
   }
 }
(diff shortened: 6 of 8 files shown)


app/app.html
```diff
@@ -15,14 +15,14 @@
       <!-- view: global_header/global_header -->
     </header>
     <div id="container" class="row">
-      <div class="row-inner">
+      <div class="row-inner-desktop-padded">
```
nchapman: Classes shouldn't contain any reference to desktop or mobile. This should all just be handled with the breakpoint code.

app/scripts/templates/global_header/global_header.html
```diff
@@ -1,4 +1,4 @@
-<div class="row-inner l-flex-horizontal">
+<div class="row-inner-padded l-flex-horizontal">
```
nchapman: This is semi unrelated, but I would prefer that we include `.l-flex-horizontal` in `.row-inner` rather than including it here. That way the CSS contains all the logic about layout rather than leaking those into the HTML. Let's discuss this more next week...

app/styles/_mixins.scss
```diff
@@ -37,3 +37,17 @@
   background-color: $background-color;
   color: $color;
 }
+
+//BREAKPOINTS
+@mixin breakpoint($point) {
```
nchapman: Nice I like this! Here's [another variation](http://www.sitepoint.com/managing-responsive-breakpoints-sass/) we could consider:

```sass
$breakpoints: (
  'small'  : ( min-width:  767px ),
  'medium' : ( min-width:  992px ),
  'large'  : ( min-width: 1200px )
);
  
@mixin respond-to($name) {
  // If the key exists in the map
  @if map-has-key($breakpoints, $name) {
    // Prints a media query based on the value
    @media #{inspect(map-get($breakpoints, $name))} {
      @content;
    }
  }
 
  // If the key doesn't exist in the map
  @else {
    @warn "Unfortunately, no value could be retrieved from `#{$breakpoint}`. "
        + "Please make sure it is defined in `$breakpoints` map.";
  }
}
```

This might be more flexible as we add more breakpoints.

app/scripts/templates/global_header/global_header.html
johngruen: @nchapman: not sure how we'd do that since the flex classes are applied in various non-row contexts and, not all rows are necessarily flexed, but yeah, I'm sure we can figure this one out.

nchapman: Something like this is what I was imagining:

```
#global-header .row-inner {
  @include .l-flex-horizontal;
}
```

johngruen: derp

app/styles/_mixins.scss
pdehaan: I also found this yesterday http://krasimirtsonev.com/blog/article/SASS-mixin-for-media-queries or this one (which is pretty close to Nick's suggestion): https://github.com/sass/sass/issues/554#issuecomment-18307019

pdehaan:
Need to poke around a bit more as to "why", but it looks like things get a bit too "edgy" at around 568px width (which seems odd to me since your breakpoint was at 520px, iirc)

(Note the left aligned Date separator and right edge of the search box getting a bit too "familiar" with the page borders)

chronicle

pdehaan:
And just for giggles, here's the page at around 350px width in the Firefox RWD simulator.
Note that the ".large" tiles show a nice full width thumbnail/screenshot, and the ".medium" tiles seem to flow nicely (although really only space for 1-2 words per row for the page titles).

chronicle

chronicle

At around 320px width things just get weird between the .medium thumbnails and the page titles with unexpected wrapping. But that's a pretty laughable breakpoint. Not sure what kind of flip-phone you'd be rocking to be browsing at that resolution.

johngruen:
@nchapman: implemented your suggested model, but it doesn't seem to be working. I gotta jet for now, but maybe there's an obvious bug i'm missing.

pdehaan:
I'm a bit suspicious of this compiled Sass->CSS code:
```css
.row-inner,
.row-inner-padded,
.row-inner-break {
  margin-left: auto;
  margin-right: auto;
  max-width: 740px;
  position: relative;
  width: 100%; }

  @media inspect() {
    .row-inner-padded {
      padding-left: 20px;
      padding-right: 20px; } }

  @media inspect() {
    .row-inner-padded {
      padding-left: 10px;
      padding-right: 10px; } }

  @media inspect() {
    .row-inner-break {
      padding-left: 20px;
      padding-right: 20px; } }
```

Me wonders if some of this specific media query+map stuff is a bit too experimental for libsass.
It looks like it's choking around this line:
```css
    @media #{inspect(map-get($breakpoints, $name))} {
      @content;
    }
```

pdehaan:
OK, last stupid comment from me, then I'll shut up...

This _MAY_ work (or at least worked for me locally):
```scss
$breakpoints: (
  small: ('max-width: 767px'),
  big: ('min-width: 768px')
);

@mixin respond-to($breakpoint) {
  @if map-has-key($breakpoints, $breakpoint) {
    @media (#{map-get($breakpoints, $breakpoint)}) {
      @content;
    }
  }
  @else {
    @warn "Unfortunately, no value could be retrieved from `#{$breakpoint}`. "
        + "Please make sure it is defined in `$breakpoints` map.";
  }
}
```
&mdash; via http://www.sitepoint.com/using-sass-maps/

--- 

### Example:
The above sauce turns this:
```scss
.row-inner {
  margin-left: auto;
  margin-right: auto;
  max-width: 740px;
  position: relative;
  width: 100%;

  &-padded {
    @extend .row-inner;
    @include respond-to(big) {
      padding-left: $double-gap;
      padding-right: $double-gap;
    }
    @include respond-to(small) {
      padding-left: $default-gap;
      padding-right: $default-gap;
    }
  }

  &-break {
    @extend .row-inner;
    @include respond-to(big) {
      padding-left: $double-gap;
      padding-right: $double-gap;
    }
  }
}
```

... into this:
```css
.row-inner, .row-inner-padded, .row-inner-break {
  margin-left: auto;
  margin-right: auto;
  max-width: 740px;
  position: relative;
  width: 100%; }

  @media (min-width: 768px) {
    .row-inner-padded {
      padding-left: 20px;
      padding-right: 20px; } }
  @media (max-width: 767px) {
    .row-inner-padded {
      padding-left: 10px;
      padding-right: 10px; } }
  @media (min-width: 768px) {
    .row-inner-break {
      padding-left: 20px;
      padding-right: 20px; } }
```

If I throw an invalid breakpoint name at it (ie: "biggie") then it seems to give me a nice warning in my Terminal:
```
Running "sass:styles" (sass) task
WARNING: Unfortunately, no value could be retrieved from `biggie`. Please make sure it is defined in `$breakpoints` map.
	 /Users/pdehaan/dev/github/chronicle/app/styles/breakpoints:17, in mixin `respond-to`
	 /Users/pdehaan/dev/github/chronicle/app/styles/modules/row:14
```

I suspect all this is just temporary and once grunt-sass updates its node-sass version to use the latest version of libsass, this all may have just worked with `inspect()` but this is probably "good enough" for now and can be revisited in the future if it causes anybody any sleepless nights.

johngruen:
@pdehaan: you are an american hero

pdehaan:
LOL! Not even American, bro! I'm just an immigrant stealing your jobs.


**Figure 1:** Represent!

johngruen:
@pdehaan: get off my property!
giphy

nchapman:
This looks great. Nicely done guys! :+1:

PR #307, opened 2015-02-09
fix(server): Removes caching for root path

### LINKED ISSUE #219: Home page isn't updating after sign in/sign out
The home page seems to often be the wrong version of the page when a user first signs in or signs out. It requires a force reload to get the proper version of the page.

nchapman:
Fixes #219.

So this feels like a hack, but it dawned on me that if we just reply with a string then hapi doesn't know the modification date and our caching problems go away. I'm ok with trashing this if there's a better way :smiley:.

@6a68 @pdehaan r?

### DIFF
--- a/server/controllers/base.js
+++ b/server/controllers/base.js
@@ -4,6 +4,7 @@
 
 'use strict';
 
+var fs = require('fs');
 var path = require('path');
 var config = require('../config');
 var STATIC_PATH = path.join(__dirname, '..', '..', config.get('server_staticPath'));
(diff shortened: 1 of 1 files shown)


6a68:
lol yeah, this is too hacky even for me ;-P

I think the right answer lies along the lines of [my comment](https://github.com/mozilla/chronicle/issues/219#issuecomment-73292024) in #219, I'm happy to look at that once the reindexing bug is done (so, the second thing on my list for this week). It's possible hapi will do stupid/helpful caching of other routes, so I should figure it out sooner than later.

nchapman:
Yeah the problem is that @pdehaan and I both messed with the options *a lot* and they didn't seem to make any difference. It seems to me those options don't work with `reply.file`. Hopefully you find the magic combination!

pdehaan:
Yeah, I was seeing the funky caching behavior with both `npm start` and running `node bin/www` locally, so I couldn't even blame it on some weird grunt-hapi issue.

PR #315, opened 2015-02-10
bug(server): change default session to one month

### LINKED ISSUE #248: I want to stay signed into Chronicle between browser restarts
Right now each time a user restarts their browser they have to sign in. This causes trouble for the add-on since it's using the same authentication.

6a68:
Fixes #248

@nchapman @pdehaan r?

### DIFF
--- a/server/config.js
+++ b/server/config.js
@@ -266,9 +266,9 @@ var conf = convict({
     env: 'SESSION_ALLOW_INSECURE_COOKIES'
   },
   server_session_duration: {
-    doc: 'Session duration in milliseconds. Zero, the default, means session cookie.',
+    doc: 'Session duration in milliseconds. Zero means session cookie. We default to one month.',
     format: 'int',
-    default: 0,
+    default: 2592000000,
     env: 'SESSION_DURATION'
   },
   server_session_cookieName: {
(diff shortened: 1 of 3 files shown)


server/config.js
```diff
@@ -266,9 +266,9 @@ var conf = convict({
     env: 'SESSION_ALLOW_INSECURE_COOKIES'
   },
   server_session_duration: {
-    doc: 'Session duration in milliseconds. Zero, the default, means session cookie.',
+    doc: 'Session duration in milliseconds. Zero means session cookie. We default to one year.',
     format: 'int',
-    default: 0,
+    default: 31556952000,
```
pdehaan: LGTM, unless you want to use a constant here, a la
```js
var SEVEN_DAYS_IN_MSEC = 1000 * 60 * 60 * 24 * 7; // 7 days
```

But the `doc` is already pretty darned descriptive.

pdehaan:
Plus see @nchapman's https://github.com/mozilla/chronicle/issues/248#issuecomment-73730118

server/config.js
6a68: can't hurt to be consistent within that file, good catch

nchapman:
Looks good! Do we want to update the `local.json.example` file as well?

6a68:
ah, yeah, sure

6a68:
alright, /me clicks!

PR #323, opened 2015-02-11
Issue 291 reindex on migrate

6a68:
@nchapman @pdehaan r?

Since we're only using one worker for now, I didn't bother integrating this into the work queue. However, I did make sure all the job variables were passed into the "worker" as the params object. Should be pretty straightforward to move this into the queue later.

Also made a couple of minor adjustments to the DB, should be self-explanatory, except the rgb -> hex conversion was added because I saw elasticsearch complaining, and might have some impact on the front-end code (if we're using the favicon colors). Note that they're stored without the leading '#'.

### DIFF
--- a/server/config.js
+++ b/server/config.js
@@ -107,6 +107,26 @@ var conf = convict({
     env: 'REDIS_DATABASE',
     format: 'int'
   },
+  db_reindex_timeout: {
+    doc: 'Timeout for an individual step of the reindexing job.',
+    format: 'int',
+    default: 30000
+  },
+  db_reindex_batchSize: {
+    doc: 'Batch size for elasticsearch reindexing',
+    default: 100,
+    format: 'int'
+  },
+  db_reindex_maxRetries: {
+    doc: 'Number of times to retry an individual reindexing step before giving up completely',
+    default: 5,
+    format: 'int'
+  },
+  db_reindex_retryWait: {
+    doc: 'Base wait period before retrying after a failure. There is a linear multiplier in the code.',
+    default: 10000,
+    format: 'int'
+  },
   // TODO put these under a 'services' namespace
   embedly_enabled: {
     doc: 'Disable embedly for running on travis.',
--- a/server/db/migrator.js
+++ b/server/db/migrator.js
@@ -5,6 +5,7 @@
 'use strict';
 
 var fs = require('fs');
+var path = require('path');
 var Joi = require('joi');
 var pg = require('pg');
 var pgpatcher = require('pg-patcher');
@@ -12,7 +13,7 @@ var q = require('q');
 
 var config = require('../config');
 var log = require('../logger')('server.db.migrator');
-var elasticsearch = require('./elasticsearch.js');
+var reindex = require('./reindex');
 
 var migrator = function (patchLevel) {
   log.debug('migrator called, patchLevel is ' + patchLevel);
@@ -41,20 +42,33 @@ var migrator = function (patchLevel) {
       log.trace(err);
       deferred.reject(err);
     } else {
-      pgpatcher(client, level, {dir: __dirname + '/migrations'}, function onPatched(err, result) {
+      pgpatcher(client, level, {dir: path.join(__dirname, 'migrations')}, function onPatched(err, result) {
         log.debug('pgpatcher callback fired');
-        // TODO: reindex elasticsearch by extracting user_pages from the migrated DB
         if (err) {
           log.error('pgpatcher migration failed: ' + err);
           done();
           pg.end();
-          deferred.reject(err);
-        } else {
-          log.info('pgpatcher migration succeeded');
+          return deferred.reject(err);
+        }
+        log.info('pgpatcher migration succeeded');
+        // at patchLevel 0 and 1, there is no user_pages table to reindex
+        if (patchLevel < 2) {
+          done();
+          pg.end();
+          return deferred.resolve(result);
+        }
+        reindex.start(function(err) {
+          if (err) {
+            log.error('elasticsearch reindexing failed: ' + err);
+            done();
+            pg.end();
+            return deferred.reject(err);
+          }
+          log.info('elasticsearch reindexing succeeded');
           done();
           pg.end();
           deferred.resolve(result);
-        }
+        });
       });
     }
   });
--- a/server/db/reindex.js
+++ b/server/db/reindex.js
@@ -1,0 +1,153 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+var q = require('q');
+var config = require('../config');
+var log = require('../logger')('server.db.reindex');
+var postgres = require('./postgres');
+var elasticsearch = require('./elasticsearch');
+var isRunning = false;
+
+// keys on the params object:
+//  done - a callback to be fired when all done, or on unrecoverable error
+//  startTime - time that reindexing started, used to ensure linear traversal through
+//              user_pages in postgres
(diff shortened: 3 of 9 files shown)


server/config.js
```diff
@@ -107,6 +107,26 @@ var conf = convict({
     env: 'REDIS_DATABASE',
     format: 'int'
   },
+  db_reindex_timeout: {
+    doc: 'Number of milliseconds for etl to fetch from postgres + bulk insert into elasticsearch',
```
6a68: ah, whoops. I initially called this 'etl' code, as in extract-transform-load. I can remove those refs from the config file.

server/config.js
```diff
@@ -107,6 +107,26 @@ var conf = convict({
...
+    format: 'int',
+    default: 30000
+  },
+  db_reindex_batchSize: {
+    doc: 'Batch size for elasticsearch reindexing',
+    default: 100,
+    format: 'int'
+  },
+  db_reindex_maxRetries: {
+    doc: 'Number of times to retry an individual ETL chunk before giving up completely',
```
6a68: ditto for `ETL` here

server/config.js
```diff
@@ -107,6 +107,26 @@ var conf = convict({
...
+    default: 5,
+    format: 'int'
+  },
+  db_reindex_retryWait: {
+    doc: 'Base wait period before retrying after a failure. There is a linear multiplier in the code.',
+    default: 10000,
+    format: 'int'
```
pdehaan: I wonder if that Convict `duration` bug got merged and released yet, if we're interested in trying to convert these to things like "10 seconds" vs "30 seconds" vs "30 days", etc.

server/db/migrator.js
```diff
@@ -43,18 +43,25 @@ var migrator = function (patchLevel) {
     } else {
       pgpatcher(client, level, {dir: __dirname + '/migrations'}, function onPatched(err, result) {
```
pdehaan: Unrelated, but `path.join(__dirname, 'migrations')` here?

server/db/reindex.js
```diff
@@ -0,0 +1,154 @@
...
+      setTimeout(function() { job(params); }, params.errorCount * params.retryWait);
+    });
+};
+
+function start(cb) {
+  log.info('starting elasticsearch reindex process');
+  if (isRunning) {
+    var msg = 'reindex is already running';
+    log.critical(msg);
+    return cb(msg);
```
pdehaan: do we want to call the callback with a string error message, or should we wrap these in an Error constructor type thing?

```js
var err = new Error('reindex is already running');
log.critical(err.message);
return cb(err);
...
```

server/embedly.js
```diff
@@ -17,6 +17,12 @@ var logFactory = require('./logger');
 var log = logFactory('server.services.embedly');
 var npmLog = logFactory('server.services.embedly.vendor');
 
+// returns hex number without the prepended '#'
+var _rgbToHex = function(r, g, b) {
+  /*jshint bitwise: false */
+  return ((1 << 24) + (r << 16) + (g << 8) + b).toString(16).slice(1);
+};
```
pdehaan: Would it be any easier to just accept an array of 3 items/colors?

```js
// returns hex number without the prepended '#'
var _rgbToHex = function(arr) {
  /*jshint bitwise: false */
  return ((1 << 24) + (arr[0] << 16) + (arr[1] << 8) + arr[2]).toString(16).slice(1);
};
```

Then you can simplify the caller by just doin':
```js
out.extractedFaviconColor = _rgbToHex(d.favicon_colors[0].color);
```

And then not have to deal with doing this:
```js
+          r = d.favicon_colors[0].color[0];
+          g = d.favicon_colors[0].color[1];
+          b = d.favicon_colors[0].color[2];
```

server/config.js
6a68: pull requests welcome

server/db/migrator.js
6a68: sure, let's do it!

6a68: going back, I also realized that we weren't cleanly handling the migration down to patch level 0, where there is no postgres `user_pages` table to extract from. Added a guard clause against that case.

server/db/reindex.js
6a68: Errors are terrible for performance, they cause the JIT to freak out. I used them here to tell Q that an individual step had been rejected, it seemed more readable than `Q.thenReject('error string')`. Not too happy with the state of signaling and flow control in this codebase right now :-\

6a68:
lol, nothing like accidentally assigning andreas when trying to pinch-zoom waffle.io on mobile. /me shakes head

server/embedly.js
6a68: yeah, great idea, that's much simpler

server/config.js
pdehaan: Sure, I can take another stab at creating a PR for that.
I think we reverted it out last time because it could be a duration in the convict file but not in the JSON files (so we'd could do "10 seconds" in the convict.js file but the JSON files needed to use 10000). I'll try retesting that locally with the latest Convict.

pdehaan: :crying_cat_face: 
https://github.com/mozilla/node-convict/issues/68

Looks like

server/db/reindex.js
```diff
@@ -0,0 +1,154 @@
...
+      log.debug('completed an iteration');
+      log.debug('result count from last iteration is ' + result);
+      // if the number of results was smaller than the stepSize, then that
+      // was the last step; we're done.
+      if (result < params.stepSize) {
+        isRunning = false;
+        log.debug('last iteration processed ' + result + ' records, step size is ' + params.stepSize);
+        log.info('reindexing complete');
+        params.done();
+        return process.exit();
```
nchapman: Exiting the process here feels wrong...

server/db/reindex.js
```diff
@@ -0,0 +1,154 @@
...
+      }, function(err) {
+        log.warn(err);
+        throw new Error('elasticsearch insertion failed');
+      });
+  };
+
+  q().timeout(config.get('db_reindex_timeout'))
+    .then(extract_)
+    .then(transform_)
+    .then(load_)
```
nchapman: These `then`s read so nicely :smile:

nchapman:
This looks great @6a68. Nicely done. I created a dummy migration and changed the `batchSize` to something less than the number of items in the test data, and I ended up with the right number of items indexed. That's a bingo. Other than my note about `process.exit()`, this looks ready to merge :+1:.

server/db/reindex.js
6a68: Yeah, good call.

server/db/reindex.js
6a68: I know, right? :-)

server/config.js
pdehaan: Filed the duration conversion issue as #327 and self assigned. Will reinvestigate and PR once the convict bug gets addressed.

6a68:
awesome, thanks for the reviews guys

6a68:
@nchapman Aha, now I remember! The migration script hangs without that `process.exit`, because the elasticsearch js library doesn't close on its own, and I was in a rush.

Taking a bit more time to look into it, it turns out elasticsearch-js doesn't document how to close a connection, or the fact that you have to close connections manually. It also doesn't have any docs whatsoever about the connection pooling which apparently is built in. Reading the source now to figure it out :-\

6a68:
Good news is, this is probably the source of whatever crashy weirdness we saw on dev when it was getting a lot of use the other week

pdehaan:
@6a68 this? https://github.com/elasticsearch/elasticsearch-js/issues/40

pdehaan:
and/or https://github.com/elasticsearch/elasticsearch-js/issues/165

6a68:
@pdehaan yeah, I was hoping to avoid linking our repo with theirs, that's why I inserted the link with spaces, not a big deal though.

PR #332, opened 2015-02-13
feat(front-end): Adds welcome with add-on installation

### LINKED ISSUE #247: I want to install the Firefox add-on
This sounds easy, but we have to figure out how to build/host/update [the add-on](https://github.com/mozilla/chronicle-addon).

nchapman:
This adds the authenticated portion of the onboarding flow at `/#welcome`. If you're a new user to Chronicle, you'll be directed here after your account is created. This is purposely unstyled and therefore ready for @johngruen to turn it into something much better. It doesn't do a whole lot for now, but will at least help you get the add-on installed in Firefox.

Fixes #247.

@6a68 @johngruen r?

### DIFF
--- a/app/scripts/lib/modal_manager.js
+++ b/app/scripts/lib/modal_manager.js
@@ -1,0 +1,160 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'underscore',
+  'jquery',
+  'velocity',
+  'velocityui'
+], function (_, $) {
+  'use strict';
+
+  function ModalManager () {
+    this._viewStates = [];
+    this.$modal = $('#modal');
+
+    // Handle default button actions
+    this.$modal.on('click', 'a.close', this.close.bind(this));
+    this.$modal.on('click', 'a.back', this.pop.bind(this));
+  }
+
+  _.extend(ModalManager.prototype, {
+    /**
+    * Opens the modal. This is a specialized version of `push` that cleans up previous views.
+    *
+    * @method open
+    * @param {Backbone.View} view Backbone view to be rendered and placed in the modal
+    * @param {Object} options Behavioral options for the modal
+    * @param {VelocityObject} options.effect Animation effect to use when opening the modal
+    * @param {Boolean} options.fullScreen Make the modal full screen
+    */
+    open: function (view, options) {
+      this._destroyViews();
+
+      this.push(view, options);
+    },
+
+    /**
+    * Pushes a view onto the modal view stack.
+    *
+    * @method push
+    * @param {Backbone.View} view Backbone view to be rendered and placed in the modal
+    * @param {Object} options Behavioral options for the modal
+    * @param {VelocityObject} options.effect Animation effect to use when opening the modal
+    * @param {Boolean} options.fullScreen Make the modal full screen
+    */
+    push: function (view, options) {
+      this._viewStates.push({ view: view, options: (options || {}) });
+
+      this._show();
+    },
+
+
+    /**
+    * Pops the top most view off the view stack.
+    *
+    * @method pop
+    */
+    pop: function () {
+      var viewState = this._viewStates.pop();
+
+      if (viewState) {
+        viewState.view.destroy();
+      }
+
+      if (this._viewStates.length > 0) {
+        this._show();
+      } else {
+        this._hide();
+      }
+    },
+
+    /**
+    * Closes the modal.
+    *
+    * @method close
+    * @param {Object} options Options for closing the modal
+    * @param {VelocityObject} options.effect Animation effect to use when closing the modal
+    */
+    close: function (options) {
+      this._destroyViews();
+      this._hide(options);
+    },
+
+
+    /**
+    * Cleans up all the views in the stack and empties the `_viewStates`.
+    *
+    * @private
+    * @method _destroyViews
+    */
+    _destroyViews: function () {
+      _.each(this.viewStates, function (viewState) {
+        viewState.view.destroy();
+      });
+
+      this._viewStates = [];
+    },
+
+    /**
+    * Shows the last view in the stack.
+    *
+    * @private
+    * @method _show
+    */
+    _show: function () {
+      var viewState = _.last(this._viewStates);
+
+      viewState.view.render();
+
+      // Force delegate events to fix an issue where restoring a previous view breaks event bindings
+      viewState.view.delegateEvents();
+
+      // TODO: Add fancy positioning: likely in the middle of the screen. We're only supporting
+      // full screen right now, so it can come later.
+
+      // Add full-screen class if the full screen option is enabled
+      if (viewState.options.fullScreen) {
+        this.$modal.addClass('full-screen');
+      } else {
+        this.$modal.removeClass('full-screen');
+      }
+
+      // Replace modal contents
+      this.$modal.html(viewState.view.el);
+
+      // Animate if an effect is provided
+      if (viewState.options.effect) {
+        this.$modal.velocity(viewState.options.effect);
+      } else {
+        this.$modal.show();
+      }
+    },
+
+    /**
+    * Hides the modal.
+    *
+    * @private
+    * @method _hide
+    * @param {Object} options Options for closing the modal
+    * @param {VelocityObject} options.effect Animation effect to use when closing the modal
+    */
+    _hide: function (options) {
+      options = options || {};
+
+      // Animate if an effect is provided
+      if (options.effect) {
+        this.$modal.velocity(options.effect, function () {
+          // Reset styles and hide
+          $(this).removeAttr('style').hide();
+        });
+      } else {
+        this.$modal.hide();
+      }
+    }
+  });
+
+  // Return a singleton
+  return new ModalManager();
+});
--- a/app/scripts/templates/welcome.html
+++ b/app/scripts/templates/welcome.html
@@ -1,0 +1,17 @@
+<h1>Welcome y'all</h1>
+
+{{#supportedBrowser}}
+<div id="install-container">
+  <p>Some details about the add-on</p>
+  <button class="install">Install the add-on</button>
+</div>
+<div id="continue-container" class="hidden">
+  <p>Go ahead and install that add-on and then continue to awesomeness.</p>
+  <button class="continue">Bring it on</button>
+</div>
+{{/supportedBrowser}}
+
+{{^supportedBrowser}}
+  <p>Sorry, there isn't currently an add-on for your browser.</p>
+  <button class="continue">Continue</button>
+{{/supportedBrowser}}
--- a/app/scripts/views/visits/index.js
+++ b/app/scripts/views/visits/index.js
@@ -15,6 +15,8 @@ define([
   'use strict';
 
   var VisitsIndexView = BaseView.extend({
+    FETCH_RETRY_WAIT: 5000,
+
     template: VisitsIndexTemplate,
 
     initialize: function () {
@@ -45,8 +47,6 @@ define([
 
     // this appends visit items rather than replacing them
     _renderVisits: function () {
-      //this.renderCollection(VisitsItemView, '.visits');
-
       var els = [];
 
       this.collection.each(function (visit) {
@@ -68,7 +68,6 @@ define([
       // let checkScrollPosition know that we're already loading more visits
       this.loadingMoreVisits = true;
 
-      // prepare the data
       var data = {
         count: this.count
       };
@@ -78,16 +77,21 @@ define([
       }
 
       // fire off the xhr request
-      var xhr = this.collection.fetch({ reset: true, data: data });
+      this.collection.fetch({ reset: true, data: data })
+        .done(this._fetchDone.bind(this))
+        .fail(this._fetchFailed.bind(this));
+    },
 
-      // keep track of the state of things
-      xhr.done(function (data) {
-        this.loadingMoreVisits = false;
+    _fetchDone: function (data) {
+      this.loadingMoreVisits = false;
 
-        // if the length is less than the count we know there aren't more results
-        this.hasMoreVisits = this.collection.length === this.count;
-        this.lastVisitId = this.collection.last().get('id');
-      }.bind(this));
+      // if the length is less than the count we know there aren't more results
+      this.hasMoreVisits = this.collection.length === this.count;
+      this.lastVisitId = this.collection.last().get('id');
+    },
+
+    _fetchFailed: function (xhr) {
+      // TODO: Do something interesting with 404s (no results)
     },
 
     _checkScrollPosition: function () {
--- a/app/scripts/views/welcome.js
+++ b/app/scripts/views/welcome.js
@@ -1,0 +1,54 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'underscore',
+  'backbone',
+  'jquery',
+  'views/base',
+  'stache!templates/welcome',
+  'lib/modal_manager',
+  'config'
+], function (_, Backbone, $, BaseView, WelcomeTemplate, modalManager, config) {
+  'use strict';
+
+  var WelcomeView = BaseView.extend({
+    className: 'welcome',
+    template: WelcomeTemplate,
+
+    events: {
+      'click button.install': 'install',
+      'click button.continue': 'continue'
+    },
+
+    install: function (event) {
+      event.preventDefault();
+
+      $.Velocity.RunSequence([
+        { e: this.$('#install-container'), p: 'transition.fadeOut' },
+        { e: this.$('#continue-container'), p: 'transition.fadeIn' }
+      ]);
+
+      // Trigger add-on installation
+      window.location = config.addon.firefox.url;
+    },
+
+    continue: function (event) {
+      event.preventDefault();
+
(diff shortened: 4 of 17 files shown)


app/scripts/views/visits/index.js
```diff
@@ -78,16 +80,24 @@ define([
...
-      }.bind(this));
+      // if the length is less than the count we know there aren't more results
+      this.hasMoreVisits = this.collection.length === this.count;
+      this.lastVisitId = this.collection.last().get('id');
+    },
+
+    _fetchFailed: function (xhr) {
+      // 404 means no results. keep retrying every NO_RESULTS_RETRY_WAIT ms
+      if (xhr.status === 404) {
+        this.fetchRetryTimer = setTimeout(this._fetch.bind(this), this.FETCH_RETRY_WAIT);
```
nchapman: I snuck this in so that new users will see visits as soon as possible. Maybe we don't want this...

app/scripts/views/welcome.js
```diff
@@ -0,0 +1,54 @@
...
+      window.location = config.addon.firefox.url;
+    },
+
+    continue: function (event) {
+      event.preventDefault();
+
+      // Set the URL to the root but don't trigger since the view is already in place
+      Backbone.history.navigate('/');
+
+      modalManager.close({ effect: 'transition.fadeOut' });
```
nchapman: This is a [Velocity UI effect](http://julian.com/research/velocity/#uiPack).

config/local.json.example
```diff
@@ -40,5 +40,6 @@
   "server_oauth_authEndpoint": "https://oauth-latest.dev.lcip.org/v1/authorization",
   "server_oauth_tokenEndpoint": "https://oauth-latest.dev.lcip.org/v1/token",
   "server_oauth_profileEndpoint": "https://latest.dev.lcip.org/profile/v1/profile",
-  "testUser_enabled": true
+  "testUser_enabled": true,
+  "addon_firefox_url": ""
```
nchapman: I'm using: `https://people.mozilla.com/~nchapman/chronicle/firefox/chronicle.xpi`

app/scripts/views/visits/index.js
```diff
@@ -41,12 +43,13 @@ define([
...
       // stop checking scroll position
       $(window).off('scroll');
+
+      // stop retrying fetch
+      clearTimeout(this.fetchRetryTimer);
     },
 
     // this appends visit items rather than replacing them
     _renderVisits: function () {
-      //this.renderCollection(VisitsItemView, '.visits');
```
nchapman: Get out of here.

app/scripts/views/visits/index.js
6a68: I'm fine with this for now, but some kind of backoff is probably best ;-)

We could prioritize "quickly processing a new user's first 50 visits" for next sprint?

We could also look into pushing updates to the client, I guess websockets or socket.io? There seem to be a couple of libraries that link socket.io messages with backbone.sync, we could use one of those or borrow their ideas and roll our own.

app/scripts/lib/modal_manager.js
```diff
@@ -0,0 +1,127 @@
...
+
+  function ModalManager () {
+    this._viewStates = [];
+    this.$modal = $('#modal');
+
+    // Listen for all clicks
+    $(document).on('click', _.bind(this._closeOnClick, this));
+
+    // Close modal on resize to prevent weirdness between desktop and mobile
+    $(document).on('resize', _.debounce(_.bind(this.close, this), 100));
```
6a68: Hmm. I wonder if flexbox gives us better odds at positioning a lightbox inside a modal vs other approaches.

This is fine for now, but it seems weird to me that you see a modal, you rotate your phone to get a better look, then it vanishes. Might be better to re-render instead of just dismiss it?

These are all details, though. I'm happy to land this as-is and polish the interactions later.

app/scripts/lib/modal_manager.js
```diff
@@ -0,0 +1,127 @@
...
+
+      this._viewStates = [];
+    },
+
+    _show: function () {
+      var viewState = _.last(this._viewStates);
+
+      viewState.view.render();
+
+      // Force delegate events to fix an issue where restoring a previous view breaks event bindings
```
6a68: Tell me more, I'm curious about this bug

app/scripts/lib/modal_manager.js
nchapman: Yeah this is left over from a different use case. I'll just remove it.

app/scripts/lib/modal_manager.js
```diff
@@ -0,0 +1,127 @@
...
+      if (viewState.options.effect) {
+        this.$modal.velocity(viewState.options.effect);
+      } else {
+        this.$modal.show();
+      }
+    },
+
+    _hide: function (options) {
+      // Animate if an effect is provided
+      if (options.effect) {
```
6a68: should we fall back to the viewState's options here? something like:

`if (options.effect || this._viewStates[0].options.effect)`?

nchapman: Unlikely since you wouldn't want something like a `fadeIn` effect when closing the modal. We'll probably have some default effects for the modals which would become the fallback.

app/scripts/lib/modal_manager.js
nchapman: It's been a while but essentially when you pull a backbone view element from the dom and then insert it again, the event bindings are broken.

app/scripts/lib/modal_manager.js
```diff
@@ -0,0 +1,127 @@
...
+          // Reset styles and hide
+          $(this).removeAttr('style').hide();
+        });
+      } else {
+        this.$modal.hide();
+      }
+    },
+
+    _closeOnClick: function (event) {
+      if (this.$modal.is(':visible')) {
```
6a68: I'm pretty sure we have a document-level listener in the main backbone js, too, so we might want to `stopPropagation` on clicks when the modal is open

app/scripts/views/visits/index.js
nchapman: Yeah I put this in here so we could experiment with what it feels like to have data loaded when the welcome experience ends. The reality is that we're going to want to do something rather fancy, and this very likely won't be the right way to do it. I'll just remove this for now.

app/scripts/lib/modal_manager.js
```diff
@@ -0,0 +1,127 @@
...
+  'velocityui'
+], function (_, $) {
+  'use strict';
+
+  function ModalManager () {
+    this._viewStates = [];
+    this.$modal = $('#modal');
+
+    // Listen for all clicks
+    $(document).on('click', _.bind(this._closeOnClick, this));
```
nchapman: Change these to `function.prototype.bind`.

app/scripts/lib/modal_manager.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  _.extend(ModalManager.prototype, {
+    open: function (view, options) {
+      this._destroyViews();
+
+      this.push(view, options);
+    },
+
+    push: function (view, options) {
```
nchapman: Document options.

app/scripts/lib/modal_manager.js
```diff
@@ -0,0 +1,127 @@
...
+      }
+
+      if (this._viewStates.length > 0) {
+        this._show();
+      } else {
+        this._hide();
+      }
+    },
+
+    close: function (options) {
```
nchapman: Document options.

app/scripts/lib/modal_manager.js
```diff
@@ -0,0 +1,127 @@
...
+        this.$modal.hide();
+      }
+    },
+
+    _closeOnClick: function (event) {
+      if (this.$modal.is(':visible')) {
+        var $target = $(event.target);
+
+        // Close if target isn't modal, modal isn't the parent, and still in the dom
+        if (!$target.is('#modal') && !$target.parents('#modal').length && $target.parents('body').length) {
```
6a68: (Sorry, I have a lot of questions - been a little while since I did front-end stuff.)

Hmm. So, "if this.modal is visible, but the target is not modal or one of its descendants, and the click is on a descendant of body, then close". I don't see how the click target could not be modal, since it's full-screen. I guess this is for the case where the modal is *not* full-screen? I guess let's just leave this code in here, although it's awkward that we don't have an underlay to provide a lightbox or whitebox effect underneath this.modal. Maybe add a todo about inserting that underlay div when we want a non-full-screen modal?

config/local.json.example
6a68: yeah, works like a charm for me!

app/scripts/templates/welcome.html
```diff
@@ -0,0 +1,17 @@
+<h1>Welcome y'all</h1>
+
+{{#supportedBrowser}}
+<div id="install-container">
+  <p>Some details about the add-on</p>
+  <button class="install">Install the add-on</button>
+</div>
+<div id="continue-container" style="display: none;">
```
pdehaan: Should we inline styles here, or use some `class="hidden"`?

We have this in the /public/styles/compiled.css:
```css
.is-hidden {
  display: none; }
```

app/scripts/views/visits/index.js
pdehaan: What happens if this is a legit 404? Or could that never happen?

app/scripts/views/welcome.js
```diff
@@ -0,0 +1,54 @@
...
+
+define([
+  'underscore',
+  'backbone',
+  'jquery',
+  'views/base',
+  'stache!templates/welcome',
+  'lib/modal_manager',
+  'config'
+], function (_, Backbone, $, BaseView, WelcomeTemplate, modalManager, config) {
```
pdehaan: `modalManager` or `ModalManager`?

app/scripts/lib/modal_manager.js
6a68: It looks like we never detach these event handlers, is that the idea behind the `:visible` check in _closeOnClick? "if we're not visible do nothing, else..." It seems like doing those visibility checks on every click will hurt performance a bit, and the code might be simpler if we just fully detach from the DOM on hide. I think we can just mark this as a todo for now and clean up later, though.

app/scripts/lib/modal_manager.js
nchapman: Yeah this doesn't really make sense in our use case (this was all borrowed from Find My Device). I was hoping to leave some of these pieces around as place holders for when we implement the rest of the modal, but I see now that they're just confusing. I'll rip out the bits we don't need and just implement them appropriately when the time comes.

app/scripts/views/welcome.js
nchapman: It's a singleton so I went with lowercase.

app/scripts/lib/modal_manager.js
nchapman: Yeah good point. I think this is going away for now so we'll figure it out when we bring it back. Likely, as you suggested, we'll have an underlay element to trap clicks and not have to rely on the document.

app/styles/modules/_global_header.scss
```diff
@@ -2,7 +2,7 @@
   @include color-scheme($white, $middle-grey);
   position: fixed;
   top: 0;
-  z-index: 100;
+  z-index: 10;
```
6a68: fine for now, but we might want to be careful about layering the z-indexes, this might be a good use case for variables, so we can see the full list of z-indexes in one place.

app/styles/modules/_modal.scss
```diff
@@ -0,0 +1,12 @@
+#modal {
+  background: $white;
+  left: 0;
+  position: absolute;
+  top: 0;
+  z-index: 20;
```
6a68: ditto for this z-index, although really, we don't have that many layers on screen right now

server/controllers/auth.js
```diff
@@ -44,7 +44,12 @@ var authController = {
       session.expiresAt = new Date(new Date().getTime() + duration);
     }
     request.auth.session.set(session);
-    reply.redirect('/');
+    if (request.auth.credentials.profile.isNewUser) {
```
6a68: lgtm, and seems to be working ^_^

app/styles/modules/_global_header.scss
nchapman: Yeah good idea. I'm hopeful we won't need much beyond what we're using now, but that would definitely help keep things sane.

app/scripts/views/visits/index.js
6a68: Good point! Honestly, the more likely recoverable error right now is a 500, we might want to just retry on failure (other than 401 unauthorized). If a user doesn't have visits yet, they'll get an empty list, not a 404, and we might want logic to trigger a retry in that case instead. @nchapman your call here

app/scripts/templates/welcome.html
nchapman: Yeah I'm always torn about the right thing to do here. Using the existing utility classes is probably a good idea.

app/scripts/views/visits/index.js
6a68: oh, whoops, just saw your "i'll remove this for now" :-)

app/scripts/views/welcome.js
6a68: I'm pretty excited to not be relying on css for animations/transitions. :beers: to velocity.js

app/scripts/views/visits/index.js
nchapman: Yeah it is going away, but it's worth pointing out that it seems the API returns a 404 instead of an empty list. I think I'd rather it return an empty list.

app/scripts/views/visits/index.js
```diff
@@ -41,12 +43,13 @@ define([
     beforeDestroy: function () {
       // stop checking scroll position
       $(window).off('scroll');
+
+      // stop retrying fetch
+      clearTimeout(this.fetchRetryTimer);
```
pdehaan: Does the `fetchRetryTimer` always exist?
Me thinks we only set the timer if we hit the `_fetchFailed` branch. But maybe doing `clearTimeout(null || undefined)` is a noop.
```js
    _fetchFailed: function (xhr) {
      // 404 means no results. keep retrying every NO_RESULTS_RETRY_WAIT ms
      if (xhr.status === 404) {
        this.fetchRetryTimer = setTimeout(this._fetch.bind(this), this.FETCH_RETRY_WAIT);
      }
    },
```

app/scripts/views/visits/index.js
```diff
@@ -78,16 +80,24 @@ define([
...
-        this.hasMoreVisits = this.collection.length === this.count;
-        this.lastVisitId = this.collection.last().get('id');
-      }.bind(this));
+      // if the length is less than the count we know there aren't more results
+      this.hasMoreVisits = this.collection.length === this.count;
+      this.lastVisitId = this.collection.last().get('id');
+    },
+
+    _fetchFailed: function (xhr) {
+      // 404 means no results. keep retrying every NO_RESULTS_RETRY_WAIT ms
```
pdehaan: ~~`NO_RESULTS_RETRY_WAIT`~~ `FETCH_RETRY_WAIT`

app/scripts/views/welcome.js
nchapman: Yeah man me too. Velocity seems pretty great. We just have to convince @johngruen that this is the best idea :).

app/scripts/lib/modal_manager.js
6a68: cool, makes sense. funny, so many front-end bugs disappear when you just re-render on (fill in the blank) change event.

app/scripts/lib/modal_manager.js
6a68: cool

app/scripts/lib/modal_manager.js
6a68: yeah, I was kind of on the fence about suggesting that; it's nice to not have extra bits lying around, but it's also nice to just land the code and tell ourselves we'll clean it up later ;-)

6a68:
@nchapman looking good, man! nice work :-)

Honestly, I was fine with landing this as initially submitted, with the exception of making sure to `stopPropagation` on events we handle here, only because it could lead to confusing bugs later and it's a nice habit to follow.

Merge as is your wont, good sir

app/scripts/lib/modal_manager.js
```diff
@@ -0,0 +1,127 @@
...
+        this._show();
+      } else {
+        this._hide();
+      }
+    },
+
+    close: function (options) {
+      if (!options) {
+        options = {};
+      }
```
pdehaan: FWIW: We do `if (!options) {...}` here and the shortcut `options: (options || {})` on line 36. Not sure if we should standardize, or whatever. I just calls em like i sees em.

nchapman: Haha good catch @pdehaan. I just pushed a fix (right before you commented) to make this one look the same.

pdehaan: I tried scanning the code, but wasn't sure if it was easier just to pass a potentially null/undefined `options` object and handle it in the `show()` and `hide()` methods  instead.
Whatever, lets get this bad boy merged!

**Figure 1:** It's go time!

nchapman: Haha. How much do we hate this move?

`options = options || {};`

Too bad we don't have Ruby's `options ||= {}` :wink:

pdehaan: I'd settle for [ES6 default params](https://github.com/esnext/es6-default-params):
```js
function add(x=0, y=0) {
  return x + y;
}
```

If this was Node, I'd almost suggest we use something like [6-to-5](https://6to5.org/) (via [**grunt-6to5**](https://github.com/6to5/grunt-6to5)), not sure if there is a decent browserify solution or polyfill.

PR #337, opened 2015-02-13
bug(search): include title, extractedLead, extractedUrl, and extractedProviderUrl when searching

### LINKED ISSUE #334: include extractedUrl, extractedProviderUrl, and title when searching?
Per https://github.com/mozilla/chronicle/pull/329#discussion_r24628825 and #324:

> _Unrelated, but should we also be searching `extractedUrl`, `extractedProviderUrl`, and/or `title`?_

> _Basically, "title" because the `extractedTitle` will be less useful if the scraped page was behind a login (ie: facebook, or something on localhost that embedly can't touch)._

@nchapman said _"Good points. Let's do a follow up PR for these. @pdehaan will you create an issue?"_

## THIS IS THAT ISSUE!!!

pdehaan:
Closes #334

### DIFF
--- a/server/models/visits.js
+++ b/server/models/visits.js
@@ -81,11 +81,15 @@ var visits = {
                 fuzziness: 'AUTO',
                 operator: 'and',
                 fields: [
+                  'title',
                   'extractedTitle',
                   'extractedContent',
+                  'extractedLead',
                   'extractedDescription',
+                  'extractedUrl',
                   'extractedProviderDisplay',
                   'extractedProviderName',
+                  'extractedProviderUrl',
                   'extractedAuthorName'
                 ]
               }
--- a/server/models/visits.js
+++ b/server/models/visits.js
@@ -89,7 +89,6 @@ var visits = {
                   'extractedUrl',
                   'extractedProviderDisplay',
                   'extractedProviderName',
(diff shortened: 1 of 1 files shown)


server/models/visits.js
```diff
@@ -81,11 +81,15 @@ var visits = {
                 fuzziness: 'AUTO',
                 operator: 'and',
                 fields: [
+                  'title',
                   'extractedTitle',
                   'extractedContent',
+                  'extractedLead',
                   'extractedDescription',
+                  'extractedUrl',
```
nchapman: The usefulness of this will really depend on tokenizer being used for this field.

server/models/visits.js
```diff
@@ -81,11 +81,15 @@ var visits = {
...
                   'extractedProviderDisplay',
                   'extractedProviderName',
+                  'extractedProviderUrl',
```
nchapman: This is probably a substring of `extractedUrl`. Let's drop this one.

pdehaan: Looks like you're right: <http://localhost:9200/chronicle/_search?_source=extractedUrl,extractedProviderUrl> (or at least for my local default dataset).

Issue #322, opened 2015-02-11
git pre-commit linter should only run on files in the git index


6a68:
The linter is slow because it lints every file in the repo, every time. The point of the git commit hook is to avoid checking in code with syntax errors. So, we should get the list of about-to-be-committed files from git (aka the git index), and only run the linter on those.

pdehaan:
Not sure if we could get something like [**grunt-newer**](https://github.com/tschaub/grunt-newer) to do this for us for the precommit hook, and then have Travis lint everything, just for giggles.

6a68:
Maybe you can use git-ls-files to get the file list and pass that to the lint task

6a68:
Ah, nope. The docs in pro git specifically mention using git status --porcelain as a stable version-independent script-friendly listing.

http://unix.stackexchange.com/questions/155046/determine-if-git-working-directory-is-clean-from-a-script

pdehaan:
Quick test w/ **grunt-newer**:
```sh
$ grunt newer:lint
Running "newer:lint" (newer) task
Fatal error: The "newer" prefix is not supported for aliases
```

Possible workaround... create a second "lint" task which explicitly calls `newer:` on each of the subtasks:
```js
  grunt.registerTask('quicklint', [
    'newer:jshint',
    'newer:jscs',
    'newer:jsonlint',
    'newer:copyright'
  ]);
```

Initial testing shows that its about 25% faster (-1 second) than calling `grunt lint`. So I *think* all we'd need to do is change the `lint` npm script from `grunt lint` to `grunt quicklint` and then Travis still runs `grunt lint` for each checkin.

@6a68 Let me know if this sounds do-able, and I can quickly PR it and you can test drive it and see if it is any better for you.
Or if this all sounds lame, we can reconsider if the precommit hook is useful -- or if we should see if we can find a solution that uses `pre-push` hook instead so the linting doesn't happen on every commit.

6a68:
@pdehaan I was hoping for 99% faster ;-)

How about a counter proposal? There's two steps here, getting the file list from git, and passing it to grunt. I looked into this a little bit:

### getting list of files from git

Here's an example of some things that could happen to a repo:

- Ask git to move a file (auto-adds it into the index): `git mv README.md README.mdz`
- Create a new file and add to the index: `touch foo && git add foo`
- Ask git to delete a file (again, auto-adds the change): `git rm server/bell-oauth-profile.js`
[...]

PR #339, opened 2015-02-14
chore(build): replace precommit-hook with husky for auto linting

pdehaan:
Replacing **precommit-hook** module with [**husky**](https://www.npmjs.com/package/husky) so we can remove some sketchy automated JSHint logic included in **precommit-hook** and convert the hook from "pre-commit" to "pre-push". See https://github.com/mozilla/chronicle/issues/322 for some more context.

This doesn't address the core issue of #322 (_"git pre-commit linter should only run on files in the git index"_), but should give a slight performance bump (if you like to `git commit` a lot and `git push` a little).

### How do I even?

To install/test this, I recommend cleaning out any existing git hooks in your ./.git/hooks/ directory that may have been set up by **precommit-hook** _before_ running `npm install`. The **husky** npm install will recreate any hooks that it watches (notably "precommit", "prepush", and "postmerge").

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -160,7 +160,7 @@
               "dependencies": {
                 "ansi-regex": {
                   "version": "0.2.1",
-                  "from": "ansi-regex@^0.2.0",
+                  "from": "ansi-regex@^0.2.1",
                   "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-0.2.1.tgz"
                 }
               }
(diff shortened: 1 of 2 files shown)


6a68:
@pdehaan works for me! I guess I can always `grunt lint` if I want to verify syntax before pushing a given commit. Please rebase, then treat yo self by hitting that green button :-)

pdehaan:
Rebased and Travis says we're green. Treating myself!

PR #340, opened 2015-02-14
fix(style): single row items are too tall

### LINKED ISSUE #297: Single row items are too tall
screen shot 2015-02-06 at 10 20 18 am

These should ~10px of padding top and bottom rather than ~20px.

johngruen:
Fixes #297

### DIFF
--- a/app/styles/modules/_user_page.scss
+++ b/app/styles/modules/_user_page.scss
@@ -82,8 +82,15 @@
       }
     }
 
+    /*
+    * tweaked position to give better
+    * alignment w/ favicon
+    */
+
     .title {
       font-size: $default-font-size - 1;
+      position: relative;
+      top: 1px;
     }
   }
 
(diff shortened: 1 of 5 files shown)


johngruen:
@pdehaan: not sure why that dump.rdb file is hanging around...added it to my global .gitignore, but it's not being very polite about not showing up in the commit as a deleted.

johngruen:
#297

app/styles/modules/_user_page.scss
```diff
@@ -103,10 +110,15 @@
...
       margin: 0;
+      padding-right: $default-gap;
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
     }
   }
 
   // Fancy sibling selectors to achieve groupings
+  // scss-lint:disable MergeableSelector
```
pdehaan: Nice trick! :clap:

pdehaan:
chronicle
**Before:** .small internal height is 104px.

chronicle
**After:** .small internal height is 76px.

PR #341, opened 2015-02-14
 style(dates): make date markers pretty

johngruen:
This should get merged after #340

### DIFF
--- a/app/styles/_base.scss
+++ b/app/styles/_base.scss
@@ -10,6 +10,7 @@
 a {
   color: $primary;
   text-decoration: none;
+  
   &:hover {
     text-decoration: underline;  
   }
--- a/app/styles/_layout.scss
+++ b/app/styles/_layout.scss
@@ -5,7 +5,7 @@ body {
 }
 
 #stage {
-  margin-top: $octo-gap;
+  margin-top: $double-grid;
 }
 
 #container {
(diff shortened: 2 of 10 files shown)


pdehaan:
chronicle
**Figure 1:** The bottom of the date marker is a bit touchy on the first history item.

johngruen:
@pdehaan gap issue fixed!

pdehaan:
Much nicer!
I'll need to tweak my local data some more since it seems to enter all the visits for the same day, but I did notice a super edge case where you spend 4 minutes deleting all your visits, the date header still remains. But we can file that as a separate issue and handle it later.

chronicle
**Figure 2:** Date header remains after you delete all your visits.

pdehaan:
Possibly less edge case? I hacked my ./bin/create_test_data.js to create data for random days and then I deleted items and it looks roughly like this. Actually, probably the same issue as my previous comment, just funnier results with multiple stacked day labels:

chronicle
**Figure 3:** January 4th January 3rd January 2nd

nchapman:
Hooray! :+1:

nchapman:
> Date header remains after you delete all your visits.

@pdehaan will you create a new issue for this?

pdehaan:
@nchapman Filed as #350

PR #344, opened 2015-02-17
chore(sass): Add a convict flag to specify Sass output style

### LINKED ISSUE #301: Add convict config for Sass outputStyle config?
Currently, debugging Sass/CSS can be a bit tricky because we hardcode the `outputStyle` to "compressed" in [/grunttasks/sass.js:11](https://github.com/mozilla/chronicle/blob/36259b99f8412c86e59c3fe2fa66f827b45147e4/grunttasks/sass.js#L11), so I'm always staring at minified content. It may be nice to have a convict config for that that defaults to "compressed" but can be overriden in my local.json file. (possible values are "nested" or "compressed" per [**grunt-sass** docs](https://github.com/sindresorhus/grunt-sass#outputstyle)).

Or this is stupid over-optimization for a extreme edge case and not worth cluttering [/server/config.js](https://github.com/mozilla/chronicle/blob/36259b99f8412c86e59c3fe2fa66f827b45147e4/server/config.js).

pdehaan:
Added a useless knob to control your Sass `outputStyle`. Possibly only useful to me and @johngruen (or anybody else dabbling in CSS who doesn't want to read minimized code).

The output is still compressed/minimized by default, but you can easily override locally by setting the `"sass_outputStyle": "nested"` in your config/local.json file.

Fixes #301

### DIFF
--- a/config/local.json.example
+++ b/config/local.json.example
@@ -41,5 +41,6 @@
   "server_oauth_tokenEndpoint": "https://oauth-latest.dev.lcip.org/v1/token",
   "server_oauth_profileEndpoint": "https://latest.dev.lcip.org/profile/v1/profile",
   "testUser_enabled": true,
-  "addon_firefox_url": ""
+  "addon_firefox_url": "",
+  "sass_outputStyle": "nested"
 }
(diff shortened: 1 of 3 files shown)


johngruen:
@pdehaan: This looks great to me. i'll let @nchapman make final rulings on that require piece. Thanks brah!

PR #346, opened 2015-02-17
refactor(db): Change the test data time interval so we see more date headings

### LINKED ISSUE #345: Chronicle test data should create visits over multiple days
Found during testing #341.
From [/bin/create_test_data.js:35-36](https://github.com/mozilla/chronicle/blob/6f3b6489250b177fe68cebdca33691a9a333939e/bin/create_test_data.js#L35-36):

> _We'll use this date as a starting point for generating records. Each successive record will be a number of seconds in the future._

> ```js
visitedAt: new Date(historyDate.getTime() + (1000 * 60 * n)).toJSON()
```

Basically all the test results are loaded into the same day. We should probably bump it up to a handful of results for each day so we can test headers and deleting a bit easier.

pdehaan:
Each successive test data visit should be +5 hours later than the previous one, giving us 5-6 entries per date heading.

FWIW, 5 [hours] is a magically random number that I made up with no significance, apart from using 2, 3, 4, 6, or 8 were all divisible by 24, giving us the exact same numbers of entries per day. Or we could increase this to an equally arbitrary 7 or 9 hours difference, if that floats your collective boats.

Fixes #345

### DIFF
--- a/bin/create_test_data.js
+++ b/bin/create_test_data.js
@@ -14,6 +14,8 @@ var user = require('../server/models/user');
 var visitsController = require('../server/controllers/visits');
 var testUrls = require('../config/test-urls');
 
+var HOURS_IN_MS = 1000 * 60 * 60;
+
 if (!config.get('testUser_enabled')) {
   throw new Error('To create test data, you must set testUser.enabled in the config.');
 }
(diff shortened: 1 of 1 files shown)


pdehaan:
r? @6a68 &mdash; I could use some help with this. It looks so obvious that it hurts, but if you clone+run this and then do the `./bin/create_db.sh` > `./bin/migrate.js` > `./bin/create_test_data.js` dance, it seems to give me most unexpected headings. For example, I didn't know there was a January 0th, but I seem to have that heading. Not sure if I messed something up, or if something else is afoot.

chronicle

**Figure 1:** What the hell damn guy?!

pdehaan:
Date heading bug fixed in #349

6a68:
sure, let's do it

PR #354, opened 2015-02-19
chore(build): make linting quicker when committing code

### LINKED ISSUE #322: git pre-commit linter should only run on files in the git index

pdehaan:
Speeds up the build process by using **grunt-newer** to only lint _modified_ files during the **pre-push** git hook (via `$ grunt quicklint`).

Fixes #322


pdehaan:
Travis seems to have failed on the tests, not my changes (which wouldn't even be exercised by Travis since it runs the full `grunt lint` now that I think about it).

6a68:
awesome! looks like the build step went just fine, which is the only thing this change could have broken on Travis.

Issue #222, opened 2015-01-28
iterate on API design


6a68:
- API endpoints seem sane enough, fine for now
- datatypes need to be revised to account for embedly stuff
  - todo: which visits fields to return in response?
  - todo: which userPage fields to return in `search` response?
- revise + standardize the response packet format, cleanly separating metadata and data
- seems necessary to add a view abstraction to transform the model data into the expected format
- use Joi schema to formalize the API-client request/response formats
- optional but nice: pubsub to connect these components (seems unlikely for this sprint)

-----
related:

nick's erb snippet from v0 - not wholly applicable (some bits correspond to features we don't/won't have), but good food for thought re: view transforms of model data ^_^

```erb
# Pastebin TehQlvGn
<div class="page <%= user_page.interesting? ? 'interesting' : 'boring' %>">
  <%= link_to title_url do %>
    <span class="image <%= user_page.image_size_ratio <= 1 ? 'landscape' : 'portrait' %>" style="background-image: url(<%= user_page.image_url %>)">
      <% if user_page.watchable? %>
        <%= icon('play') %>
      <% end %>
    </span>
  <% end %>
  <div class="text">
    <div class="provider">
      <%= image_tag(user_page.favicon_url(image_url('globe.png')), class: 'favicon', width: 16, height: 16) if user_page.favicon_url %>
      <%= link_to user_page.provider_name, user_page.provider_url, class: 'name' %>
      <span class="author"><%= user_page.author_name %></span>
    </div>
    <h3 class="title"><%= link_to(truncate(user_page.title || user_page.url, length: 90), title_url) %></h3>
    <div class="url"><%= link_to user_page.url %></div>
    <div class="description"><%= first_sentence(user_page.summary) %></div>
    <div class="actions">
      <% if user_page.watchable? %>
        <%= action_button user_page, :saved, label: 'Watch Later', active_label: 'Watching Later', class: 'save', icon: 'mail-forward' %>
      <% elsif user_page.readable? %>
        <%= action_button user_page, :saved, label: 'Read Later', active_label: 'Reading Later', class: 'save', icon: 'mail-forward' %>
      <% end %>
      <%= action_button user_page, :liked, label: 'Like', active_label: 'Liked', class: 'like', icon: 'heart' %>
      <%= action_button user_page, :archived, label: 'Archive', active_label: 'Archived', class: 'archive', icon: 'check' %>
    </div>
  </div>
</div>
```

peter's list of useful fields from goofing around with embedly locally:
[...]

PR #355, opened 2015-02-19
feat(docs): add a /docs route using Lout

### LINKED ISSUE #108: Use Lout for route docs?
Not sure if this is useful or not (but may be handy if running in development/local dev mode). I don't know that I'm comfortable with this going into Prod as an endpoint (not sure why, it's easy to see the endpoints in source anyways), but yeah, we can probably put a handy switch and only include Lout if your `config.get('env')` is "local", "awsdev", or "test" in [`NODE_ENV`](https://github.com/mozilla/chronicle/blob/9767c8d0d595d98cf06e4ecb884951d543328b83/server/config.js#L13-L18) -- or just add a Boolean config flag for "showApiDocs".

Just for giggles, I did an `npm i lout -D` and then added the [Lout](https://github.com/hapijs/lout) plugin to our `server.register()` call (where `Lout` was `var Lout = require('lout');`'ed earlier):

```js
server.register([
  HapiAuthCookie,
  Bell,
  Lout
], function (err) {
```

Now when I goes to http://localhost:8080/docs I see the following sauce:
documentation_for_rest_routes
**Figure 1:** Route list

documentation_for
**Figure 2:** Route details

pdehaan:
Adds a <http://localhost:8080/docs> route which shows some auto-generated route documentation using [**Lout**](https://www.npmjs.com/package/lout).

It ain't pretty, but _meh_.

Closes #108

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -7,6 +7,7 @@
 var Hapi = require('hapi');
 var HapiAuthCookie = require('hapi-auth-cookie');
 var Bell = require('bell');
+var Lout = require('lout');
 var pg = require('pg');
 
 var config = require('./config');
@@ -37,7 +38,8 @@ server.connection({
 
 server.register([
   HapiAuthCookie,
-  Bell
+  Bell,
+  Lout
 ], function (err) {
   if (err) {
     log.warn('failed to load plugin: ' + err);
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -160,7 +160,7 @@
               "dependencies": {
                 "ansi-regex": {
                   "version": "0.2.1",
-                  "from": "ansi-regex@^0.2.0",
+                  "from": "ansi-regex@^0.2.1",
                   "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-0.2.1.tgz"
                 }
               }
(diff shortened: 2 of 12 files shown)


server/index.js
```diff
@@ -37,7 +38,8 @@ server.connection({
 
 server.register([
   HapiAuthCookie,
-  Bell
+  Bell,
+  Lout
```
pdehaan: Not sure about this bit right here.
I don't know if we should just blindly include Lout always, or if we want to wrap this in a `config.get('env')` and only include the /docs route if we're not in "production" environment.

pdehaan:
r? @nchapman so he can look at how stupid this stupid is, and say "this is stupid, i don't want this" and close this.
Or merge it, whatever. Not like I'm paid per pull-request.

But yeah, I'm sure we could use some better descriptions, but figure we'll do baby steps and see if it's worth proceeding further before I try and copy/pasta [/docs/API.md](https://github.com/mozilla/chronicle/blob/master/docs/API.md) into various descriptions.

pdehaan:
documentation_for_rest_routes

server/index.js
6a68: I think it's fine to include it, and probably fine to load the lout route in production, too: our code's open source, we aren't really exposing anything.

6a68:
@pdehaan I'm game to try this, let's land it. I'm going to do a cleanup pass at the API this sprint (#222), seems like a good opportunity to give this a shot.

6a68:
Ah, looks like you'll need to rebase before landing. (If you want, you can start pushing branches to mozilla instead of your fork, then I can rebase-and-merge stuff like this for you.)

pdehaan:
Rebased. Travis green. Self merge.

PR #357, opened 2015-02-23
feat(server): restrict access to whitelisted users

### LINKED ISSUE #128: I want to join chronicle with a valid mozilla.com email address

6a68:
Currently returning a 401 with a nice error string, we could do more.

Fixes #128.

### DIFF
--- a/config/local.json.example
+++ b/config/local.json.example
@@ -1,4 +1,5 @@
 {
+  "alpha_allowedUsers": ["your@emailhere.com"],
   "db_elasticsearch_host": "127.0.0.1",
   "db_elasticsearch_port": 9200,
   "db_elasticsearch_queryTimeout": 15000,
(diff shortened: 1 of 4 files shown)


6a68:
Current error looks like:

screenshot 2015-02-23 09 17 27

We could certainly do more, but this works for now.

6a68:
@nchapman r?

@pdehaan The new lint-before-push code seemed to work quite nicely ^_^

6a68:
Note that this pull request does not specify what set of emails we are allowing for right now, but not using a regex does imply that it's a whitelist and not a generic "all at mozilla.com" or whatever. When we get to a "friends and family" beta release, we can update this to allow whole domains, if we want.

6a68:
Not sure if travis is perma-red right now, giving it a second kick

pdehaan:
Did we change something in Travis recently? What's up with the two checks?

6a68:
I have no idea; the .travis.yml file hasn't been edited since the 6th, and I'm pretty sure I haven't seen that 2 checks thing until today.

PR #358, opened 2015-02-23
docs(signoff): create Chronicle Prerelease Deployment Checklist

pdehaan:
Converted from a wiki page to a markdown file in /docs/.

### DIFF
--- a/docs/signoff.md
+++ b/docs/signoff.md
@@ -1,0 +1,41 @@
+# Chronicle Prerelease Deployment Checklist
+
+Before we hand a build off to ops, we should perform the following checks (at a minimum) to make sure the app is ready to be deployed.
+
+## General:
+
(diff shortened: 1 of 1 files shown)


6a68:
@pdehaan This is great, thanks! I guess we should make it clear that the dependency check should happen before checking that the features work, maybe make this one huge checklist? I'm not sure, do what seems best. Feel free to check it in when you're satisfied (which might be right now).

Also, this is something we can address in a follow-on issue, but: what could we do to more fully automate this checklist? I know we'd need additional intern tests to check features are working, but maybe it'd be a quick win to add the additional scss and html grunt tasks (if those actually do catch errors), and create an npm task to run through the automatable parts of the checklist.

pdehaan:
I find that the _scss-lint_, _csslint_, and _htmllint_ tasks usually have more false positives than legit issues. But yeah, we could turn them into real-live Grunt tasks and commit into the repo and then bury them behind some "grunt signoff-check" task which we only ever run manually.

Issue #257, opened 2015-02-03
Scraper should discard embedly results with type:error


pdehaan:
Not sure what flow is happening here. I'm visiting http://pto.mozilla.org/ (needs LDAP password, but I'm logged in), but Chronicle displays the title as "401 Authorization Required".

Scraping the page with Embedly (web) gives me a rather unpleasant red bar:
http://embed.ly/docs/explore/extract?url=http%3A%2F%2Fpto.mozilla.org%2F

> _The URL you entered is not supported by embedly. If you think it should be, tell us here._

But what I'm not understanding is why, if the page title is being returned by the Firefox add-on, and the actual page title (in Firefox) is "Mozilla PTO", where is the title being set as "401 Authorization Required".

It definitely LOOKS like that 401 is coming from Embed.ly. Here's the [neutered] output from my embedly CLI tool:

```json
{
  "cache_age": 80525,
  "content": null,
  "description": null,
  "error_code": 401,
  "error_message": "HTTP 401: Unauthorized",
  "lead": null,
  "original_url": "http://pto.mozilla.org",
  "provider_display": "intranet.mozilla.org",
  "provider_name": "Mozilla",
  "provider_url": "https://intranet.mozilla.org",
  "published": null,
  "safe": true,
  "title": null,
  "type": "error",
  "url": "https://intranet.mozilla.org/pto/"
}
```

6a68:
Nice catch, @pdehaan! Once we get the happy paths sketched out more completely, it'll be great to add some logic to the scraper to handle common error cases like this.

6a68:
not clear if type:error is returned when embedly itself times out, or if it's just a 401

pdehaan:
Yeah, it would probably be useful to log all the errors and just not display them on the front end so we can audit the errors and see if we can find trends.
Here's a snippet from my localhost w/ sample data and an "error" `extractedType`:

```js
{
  _index: "chronicle",
  _type: "userPages",
  _id: "98f605d0-dd25-4bf8-bf57-124921564fb5",
  _score: 1,
  _source: {
    extractedCacheAge: 79608,
    extractedContent: null,
    extractedDescription: null,
    extractedFaviconUrl: null,
    extractedLanguage: null,
    extractedLead: null,
    extractedProviderDisplay: "sfbay.craigslist.org",
    extractedProviderName: "Craigslist",
    extractedProviderUrl: "http://sfbay.craigslist.org",
    extractedSafe: true,
    extractedTitle: null,
[...]

PR #359, opened 2015-02-24
WIP - Issue 330 store all embedly data

6a68:
@pdehaan mind taking a look? This is not ready to land; I need to update the controller/view layer to fix the API, but there are so many fiddly little changes here, and I'd really appreciate another set of eyes on this.

Previously, I was making things unnecessarily complex at the DB layer:

We were only storing a subset of the embedly response, doing unnecessary transforms on that response. Now, we store the whole thing using the postgres native json type. Queries get so much simpler, and embedly.js gets so much simpler.

All the keys in each record were underscored in postgres, but camelCased in elasticsearch. Also, we were structuring the data slightly differently in both, by nesting certain fields one level deep in elasticsearch, but not in postgres. Now, we keep the same structure in elasticsearch and postgres, where the embedly data is under an extracted_data key, and everything else is top level. This makes the pg -> es mapping super super simple, which is good, since we do it in several places (visit.create, user-page.update). There is one added bit of complexity: where the postgres db object used to always camelize its results, now we only sometimes camelize them. I think this can be fixed, but not right now. Another added bit of complexity is that postgres.camelize now has to separately camelize the contents of the embedly response; I opted for dumb nested `Object.keys().forEach` calls, rather than fancy recursion. I think it's more readable; we'll reassess if we go another level deeper.

Other random notes:

Once AWS adds 9.4 support, we can consider moving to the jsonb type. This wouldn't bring us any significant improvements right now anyway, since we don't do any querying on the contents of the embedly response, but jsonb auto-indexes all the keys, and converts json to an internal binary representation on insert - so it would presumably make it easy and fast to include those keys in queries. Seems like it would give us really cool flexibility with our data.

I haven't added anything into a user-page record to indicate which scraper scraped the page, because we only have one scraper right now. Keeping it simple, minimizing the amount of code.

I've added the error check inside embedly.js, so we can avoid storing responses with the error type (#257).


server/db/postgres.js
```diff
@@ -33,6 +33,14 @@ var postgres = {
...
       Object.keys(row).forEach(function(k) {
         output[camelize(k)] = row[k];
+        // special case: json extracted_data blob
+        // rather than use recursion, take advantage of the fact that all underscored keys
+        // are top-level and just nest another function. stupid, simple, works.
+        if (k === 'extracted_data') {
+          Object.keys(row.extracted_data).forEach(function(key) {
+            output.extractedData[camelize(key)] = row.extracted_data[key];
+          });
+        }
```
pdehaan: i'm a bit unclear here... We aren't using `k` in the inner `forEach()` or anything, should we just move this whole block outside of the outer `Object.keys(row).forEach()` call so we don't check the `if` statement each time?

server/db/migrations/patch-2-3.sql
```diff
@@ -1,2 +0,0 @@
--- used for batch reindexing based on creation time
-CREATE INDEX user_pages_created_at ON user_pages (created_at);
```
pdehaan: Wait, can we delete migrations? Or don't we care yet since we aren't in prod and this is all just local and we're probably deleting the database each time anyways?

6a68: you guessed it! I'm avoiding the horror of "hmm, how to migrate the data" as long as possible

server/db/postgres.js
6a68: good point, it doesn't really need to be nested. I'll fix.

6a68: fixed

pdehaan:
I'm struggling to reset my database with this PR.

I'm in ./bin/ and running `$ ./create_db.sh` and getting the following output:
```sh
DROP DATABASE
REASSIGN OWNED
DROP ROLE
CREATE ROLE
CREATE DATABASE
GRANT
ALTER SCHEMA
curl: (7) Failed to connect to localhost port 9200: Connection refused
curl: (7) Failed to connect to localhost port 9200: Connection refused
```

Not sure why Elastic is failing to connect, I see this if I go to <http://localhost:9200> directly:
```json
{
  "status" : 200,
  "name" : "Masked Marauder",
  "cluster_name" : "elasticsearch_pdehaan",
  "version" : {
    "number" : "1.4.2",
    "build_hash" : "927caff6f05403e936c20bf4529f144f0c89fd8c",
    "build_timestamp" : "2014-12-16T14:11:12Z",
    "build_snapshot" : false,
    "lucene_version" : "4.10.2"
  },
  "tagline" : "You Know, for Search"
}
```

If I go to <http://localhost:9200/chronicle/> I get this:
```
{
  "chronicle": {
    "mappings": {
      "userPages": {
        "properties": {
          "extractedAt": {
            "type": "date",
            "format": "dateOptionalTime"
          },
          "extractedAuthorName": {
            "type": "string"
          },
          "extractedAuthorUrl": {
            "type": "string"
          },
          "extractedCacheAge": {
            "type": "long"
          },
          ...
      "user_pages": {
        "properties": {
          "created_at": {
            "type": "date",
            "format": "dateOptionalTime"
          },
          "extracted_data": {
            "properties": {
              "app_links": {
                "properties": {
                  "app_name": {
                    "type": "string"
                  },
                  "app_store_id": {
                    "type": "string"
                  },
                  "package": {
                    "type": "string"
                  },
                  "type": {
                    "type": "string"
                  },
                  "url": {
                    "type": "string"
                  }
                }
              },
              "authors": {
                "properties": {
                  "name": {
                    "type": "string"
                  },
                  "url": {
                    "type": "string"
                  }
                }
              },
              "cache_age": {
                "type": "long"
              },
```
(So basically i have `userPages` and `user_pages`)

But after doing a ./migrate.js and ./create_test_data.js, I see the following barf in my Elastic Terminal tab:
```
[2015-02-24 13:08:25,142][DEBUG][action.index             ] [Masked Marauder] [chronicle][1], node[KFZQnMQ-Rx-rMC6ulComaQ], [P], s[STARTED]: Failed to execute [index {[chronicle][user_pages][d13e811e-8db0-40bf-bdcb-e5ca7d751f57], source[_na_]}]
org.elasticsearch.index.mapper.MapperParsingException: failed to parse, document is empty
	at org.elasticsearch.index.mapper.DocumentMapper.parse(DocumentMapper.java:559)
	at org.elasticsearch.index.mapper.DocumentMapper.parse(DocumentMapper.java:490)
	at org.elasticsearch.index.shard.service.InternalIndexShard.prepareCreate(InternalIndexShard.java:392)
	at org.elasticsearch.action.index.TransportIndexAction.shardOperationOnPrimary(TransportIndexAction.java:198)
	at org.elasticsearch.action.support.replication.TransportShardReplicationOperationAction$AsyncShardOperationAction.performOnPrimary(TransportShardReplicationOperationAction.java:511)
	at org.elasticsearch.action.support.replication.TransportShardReplicationOperationAction$AsyncShardOperationAction$1.run(TransportShardReplicationOperationAction.java:419)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

pdehaan:
Same results if I run the `curl` directly:
```sh
$ curl -XDELETE 'http://localhost:9200/chronicle/'
curl: (7) Failed to connect to localhost port 9200: Connection refused
```

And no debug output appears in my Elastic tab.

server/controllers/visit.js
```diff
@@ -8,14 +8,8 @@ var Boom = require('boom');
...
 var log = require('../logger')('server.controllers.visit');
-var url2png = require('url2png')(config.get('url2png_apiKey'), config.get('url2png_secretKey'));
 var visit = require('../models/visit');
-
-function addScreenshot(item) {
-  // TODO this is the seed of a view or view helper, believe it or not
-  item.screenshot_url = url2png.buildURL(item.url, {viewport: '1024x683', thumbnail_max_width: 540});
-  return item;
-}
+var view = require('../views/visit');
```
pdehaan: Not sure if this should be something like `var visitView = ...;` for consistency with [`var visitsView = require('../views/visits');`](https://github.com/mozilla/chronicle/pull/359/files#diff-fc2554804aa3e40d22878ae59ff7baa9R15) below.

server/views/visits.js
```diff
@@ -0,0 +1,20 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+var visitView = require('./visit');
+
+var visitsView = {
+  render: function(visits) {
+    if (!visits.length) { visits = [visits]; }
```
pdehaan: What does this line do?
`if visits.length == 0, set visits to an empty array`?

server/views/visit.js
```diff
@@ -0,0 +1,37 @@
...
+
+  // remove the userId from the visit, if present
+  delete results.userId;
+
+  newResult.userPage = results;
+  return newResult;
+};
+
+var addScreenshot = function(item) {
+  item.screenshot_url = url2png.buildURL(item.url, {viewport: '1024x683', thumbnail_max_width: 540});
```
pdehaan: A bit silly, but the `1024x683` and `540` seem a bit "magic number-ish" here. Not sure if we should move them into consts at the top, or sillier still, into a convict config var.

Feel free to resume laughing deeply whilst deleting this absurd comment. #seesomethingsaysomething

server/db/postgres.js
```diff
@@ -34,6 +34,13 @@ var postgres = {
       Object.keys(row).forEach(function(k) {
         output[camelize(k)] = row[k];
       });
+      // special case: json extracted_data blob
+      if (output.extractedData) {
+        Object.keys(output.extractedData).forEach(function(key) {
+          output.extractedData[camelize(key)] = output.extractedData[key];
+          delete output.extractedData[key];
```
6a68: derp. if the key is 'images' and doesn't contain underscores, this will delete it

server/views/visits.js
6a68: hmm. Is this any clearer? `if (!Array.isArray(visits)) { visits = [visits] }`? "if it's not an array, wrap it in an array."

server/views/visit.js
6a68: lol, yeah, I think I'll let it look obviously wrong until we start proxying those instead of doing the super-slo-mo API request on the client.

server/controllers/visit.js
6a68: good call, I was on the fence about which one to use, but never made up my mind

6a68:
@pdehaan That's weird: it sounds like you can GET `localhost:9200` just fine, but when you try to send it a DELETE, you can't even connect. Maybe try restarting the server? See if you can GET via curl instead of via the browser. Weird.

6a68:
gotta love a squash + rebase! I'll exercise this some more tomorrow, reread it, and land it. Comments welcome.

6a68:
Everything seems to be working on localhost for me, so I'm going to merge this. I found a bug that I'll fix in a separate followup commit (#362).

6a68:
Looks like travis has been sitting for 16 hours but never actually built the updated branch from last night. Hopefully it's just a transient bug and not something that happens again.

PR #368, opened 2015-04-16
bug(db): specify TEMPLATE when creating db.

### LINKED ISSUE #367: db create script: specify template to avoid collation errors
If the user is on a system where `en_US.UTF-8` isn't the default collation, attempting to create_db will fail with an error like:

```
ERROR:  new collation (en_US.UTF-8) is incompatible with the collation of the template
database (C)
HINT:  Use the same collation as in the template database, or use template0 as template.
```

One solution which seems to work is to specify the template from which the database should be built. So, we'd replace this line:

```
psql -c "CREATE DATABASE chronicle ENCODING 'UTF-8' LC_COLLATE = 'en_US.UTF-8' 
  LC_CTYPE = 'en_US.UTF-8';" -U $PSQLUSER
```

with

```
psql -c "CREATE DATABASE chronicle ENCODING 'UTF-8' LC_COLLATE = 'en_US.UTF-8' 
  LC_CTYPE = 'en_US.UTF-8' TEMPLATE template0;" -U $PSQLUSER
```

Solution reference: http://stackoverflow.com/questions/18870775/how-to-change-the-template-database-collection-coding-on-postgresql

jaredhirsch:
By specifying the template, we avoid collation errors when running on clients
  that do not have the en_US locale.

  Fixes #367.

### DIFF
--- a/bin/create_db.sh
+++ b/bin/create_db.sh
@@ -15,7 +15,7 @@ psql -c "REASSIGN OWNED BY chronicle TO $PSQLUSER;" -U $PSQLUSER
 psql -c 'DROP USER IF EXISTS chronicle;' -U $PSQLUSER
 
 psql -c "CREATE USER chronicle WITH PASSWORD 'chronicle';" -U $PSQLUSER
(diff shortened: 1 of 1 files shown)


6a68:
works for me, but travis seems to be failing due to a version of postgres that doesn't support the 'json' datatype...merging

